PR #1, opened 2020-04-30
Add shortcut for searching and inspecting shell variables

patrickf3139:


patrickf3139:
Oops, used the wrong merge algorithm. Gonna revert this and re-try.

PR #3, opened 2020-04-30
Add shortcut for searching and inspecting shell variables

patrickf3139:
Ctrl+v (v for variable) to browse through all local variables using fzf with their values in a preview pane and output the selected variable name.


patrickf3139:
Already merged

Issue #19, opened 2020-08-18
Ctrl+f file preview is broken on linux


chshbh:
I think file [`__fzf_preview_file.fish`](https://github.com/patrickf3139/fzf.fish/blob/master/functions/__fzf_preview_file.fish) do not yield appropriate output for Linux systems. 

> Here it is on Gnome Manjaro.
![image](https://user-images.githubusercontent.com/9166890/90463045-a8c9e800-e127-11ea-9a74-7878b6622493.png)

patrickf3139:
Ah shoot, thanks for reporting.
What's your output for
```
file --verison
file-5.37
magic file from /usr/share/file/magic
```
?

patrickf3139:
In the meantime, I would recommend switching back to version 2.
```
fisher rm patrickf3139/fzf.fish
fisher add patrickf3139/fzf.fish@v2.0
```

chshbh:
> Ah shoot, thanks for reporting.
> What's your output for

```
file --version
file-5.39
magic file from /usr/share/file/misc/magic
seccomp support included
```

chshbh:
I fixed the issue, but I am no longer using `file`, rather `ls`. I can send a pull request, locally I am using [exa](https://github.com/ogham/exa) which is faster than `file` or `ls` command.  IMO `ls` is fine too, since differences are in milliseconds.

patrickf3139:
Yes please PR! Thank you

patrickf3139:
Fixed in #21 .

PR #20, opened 2020-08-18
removed file command dependency, file -i behaviour is not consistent across the platforms

chshbh:
`ls` is fast enough but can be substituted with `exa`. #19

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -1,17 +1,26 @@
 # helper function for __fzf_search_current_dir
 function __fzf_preview_file --description "Prints a preview for the given file based on its file type." --argument-names file_path
-    set file_type (file --brief -i "$file_path")
+    if test -d "$file_path"
+        set file_type "d"
+    else
+        set file_type (ls -o "$file_path" | cut -c 1) #exa is faster than ls and file
+    end
     switch $file_type
-        case "directory"
+        case "d"
             # Setting CLICOLOR_FORCE forces colors to be enabled even to a non-terminal output
             CLICOLOR_FORCE=true ls -a "$file_path"
-        case "socket"
+        case "s"
             __fzf_report_file_type "$file_path" "socket"
-        case "broken symbolic link*" # example: "broken symbolic link to /path/to/file""
-            __fzf_report_file_type "$file_path" "broken symbolic link"
-        case "* special *" # examples: "character special (23/5)" and "block special (1/7)"
+        case "l" # example: "symbolic link to /path/to/file"
+              #no way to directly tell if broken symlink using ls
+              if test -e (readlink -f "$file_path")
+                  __fzf_report_file_type "$file_path" "symbolic link"
+              else
+                  __fzf_report_file_type "$file_path" "broken symbolic link"
+              end
+        case "b" "c" # examples: "character special (23/5)" and "block special (1/7)"
             __fzf_report_file_type "$file_path" "device file"
-        case "regular file"
+        case "-"
             bat --style=numbers --color=always "$file_path"
         case "*"
             echo "Unsure how to preview '$file_path', which has file type '$file_type'. Please open an issue at https://github.com/patrickf3139/fzf.fish."


functions/__fzf_preview_file.fish
```diff
@@ -1,17 +1,26 @@
 # helper function for __fzf_search_current_dir
 function __fzf_preview_file --description "Prints a preview for the given file based on its file type." --argument-names file_path
-    set file_type (file --brief -i "$file_path")
+    if test -d "$file_path"
+        set file_type "d"
+    else
+        set file_type (ls -o "$file_path" | cut -c 1) #exa is faster than ls and file
```
patrickf3139, line 6: So a clean installation of exa aliases itself to ls?

functions/__fzf_preview_file.fish
```diff
@@ -1,17 +1,26 @@
 # helper function for __fzf_search_current_dir
 function __fzf_preview_file --description "Prints a preview for the given file based on its file type." --argument-names file_path
-    set file_type (file --brief -i "$file_path")
+    if test -d "$file_path"
+        set file_type "d"
```
patrickf3139, line 4: does exa not tell you if the file in question is a directory?

functions/__fzf_preview_file.fish
```diff
@@ -1,17 +1,26 @@
...
             CLICOLOR_FORCE=true ls -a "$file_path"
-        case "socket"
+        case "s"
             __fzf_report_file_type "$file_path" "socket"
-        case "broken symbolic link*" # example: "broken symbolic link to /path/to/file""
-            __fzf_report_file_type "$file_path" "broken symbolic link"
-        case "* special *" # examples: "character special (23/5)" and "block special (1/7)"
+        case "l" # example: "symbolic link to /path/to/file"
+              #no way to directly tell if broken symlink using ls
+              if test -e (readlink -f "$file_path")
```
patrickf3139, line 16: I love this, very clever. Nice work!

functions/__fzf_preview_file.fish
```diff
@@ -1,17 +1,26 @@
...
+    else
+        set file_type (ls -o "$file_path" | cut -c 1) #exa is faster than ls and file
+    end
     switch $file_type
-        case "directory"
+        case "d"
             # Setting CLICOLOR_FORCE forces colors to be enabled even to a non-terminal output
             CLICOLOR_FORCE=true ls -a "$file_path"
-        case "socket"
+        case "s"
```
patrickf3139, line 12: Good to know exa tells you if something is a socket. Thanks for trying your best to preserve the intended behavior!

functions/__fzf_preview_file.fish
chshbh: No, I am using exa on my machine and you would have to install exa, its a single binary. You would have to either:
+ replace all the instance of ls manually in the file
+ idk if aliases would work, i use global abbr, so ls is replaced in shell with exa but idk how to make it work with scripts. 

AFAIK, ls was 10x slower than file, but since file was running in micro seconds per file, that makes ls still in milli seconds.

functions/__fzf_preview_file.fish
chshbh: +Yes, in fact, this was done for ls specifically. With ls, you get a total block in first-line when using -a option, why -a? to get the file flag if the file in question is a directory. It was a choice between 2 if-else block and using sed + awk to get the file flag needed.
+In the case of exa, you have -d option, which beautiful turns quiet if its a file. 
+So, `exa -dl $file_path | cut -c 1` would suffice.

functions/__fzf_preview_file.fish
chshbh: + Both ls and exa would suffice for this.
+ [Here](https://en.wikipedia.org/wiki/Unix_file_types) have a look, if you want special use case for each type of file, its possible too.
+ You do not need exa, but if you want I can push exa file for you. exa, in general, is prettier and faster than ls.

patrickf3139:
Hey @chshbh great work, and thanks for explaining to me how your PR works! I have a few changes I'd like to make. I hope you don't mind me pushing to your branch and then merging it--don't worry, you will get the credit for it.

patrickf3139:
Ah darn, I don't have permissions to push to your branch. Can you give them to me, please? https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

chshbh:
> Ah darn, I don't have permission to push to your branch. Can you give them to me, please? https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I think you can edit my commit.

patrickf3139:
Hey @chshbh , sorry I decided to just open my own PR and fix it because this bug was live and I didn't want people to not have a working file preview for too long.
I want to thank you again for filing the issue and writing this PR, which I based my fix off off! I gave you credit in the commit.

PR #21, opened 2020-08-18
fix file preview not working non non-Mac OSes

patrickf3139:
# Problem
File preview wasn't working on Linux and presumably Windows because the `file` command that ships with Mac, used to determine the file type, is outdated and has different options. See #19

# Solution
Use `ls` to get the file type instead. Big thanks to @chshbh for the idea of using `ls`! He opened up a great PR in #20

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -24,6 +24,9 @@ function __fzf_preview_file --description "Prints a preview for the given file b
             set target_path (realpath $file_path)
             set_color yellow; and echo "'$file_path' is a symlink to '$target_path'"
             __fzf_preview_file "$target_path"
+        case "" # occurs when ls failes, e.g. with bad file descriptors
+            set_color red; and echo 'ls filed to get the file type..' >&2
+            exit 1
         case "*"
             echo "Unexpected file symbol $file_symbol. Please open an issue at https://github.com/patrickf3139/fzf.fish."
     end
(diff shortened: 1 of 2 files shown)


functions/__fzf_report_file_type.fish
```diff
@@ -1,6 +1,6 @@
 # helper function for __fzf_preview_file
 function __fzf_report_file_type --description "Explain the file type for a file." --argument-names file_path file_type
     set_color red
-    echo "Cannot preview: '$file_path' is a $file_type."
+    echo "Cannot preview '$file_path': it is a $file_type."
```
patrickf3139, line 4: This makes more grammatical sense.

functions/__fzf_preview_file.fish
```diff
@@ -1,19 +1,33 @@
 # helper function for __fzf_search_current_dir
 function __fzf_preview_file --description "Prints a preview for the given file based on its file type." --argument-names file_path
-    set file_type (file --brief -i "$file_path")
```
patrickf3139, line 3: main change: switching from using file to get file type to using ls to get file type symbol

functions/__fzf_preview_file.fish
```diff
@@ -1,19 +1,33 @@
...
-    switch $file_type
-        case "directory"
```
patrickf3139, line 5: aAlso rearranged cases into two groups: happy paths and unhappy paths.

Issue #28, opened 2020-09-12
feature request : bindings for abbrs and functions


matthiasdebernardini:
A quick hotkey for looking at your abbrs and functions.

this here works for me pretty well, would this be hard to add or out of scope for the project?

I currently have the command below abbreviated to `gabb` for get abbr. 

`bat .config/fish/conf.d/abbreviations.fish | rg abbr | fzf`

patrickf3139:
Hi Mattias, I like your creativity! I am unfamiliar with fish abbreviations. Are they always stored in `.config/fish/conf.d/abbreviations.fish`?

matthiasdebernardini:
Hi Patrick,

Thanks!

Fish abbreviations are essentially aliases done the fish way.

An alias can be 
`alias gc="git checkout"`
whereas an abbr can be
`abbr gc "git checkout"`

However, abbr's substitute "in place" so your history is a bit cleaner and IMO is a bit more clear whats going on, so it's easier to read.

They are not always stored in this file name, so I should probably change that. 

Maybe I can grep for every instance of abbr in every file that ends with `.fish`

This could also be used to look at aliases, since I usually forget them when I am trying to learn them.

with regards to the interoperability, do you have fd installed on your Mac?

I am using fd-find from rust (because its faster and has simpler syntax)

however I don't know how this would work on windows

patrickf3139:
Hi Matt,

> They are not always stored in this file name, so I should probably change that.
Maybe I can grep for every instance of abbr in every file that ends with .fish

That's a lot better than assuming the user always stores it in that exact file. However, I still think that's a non-starter due to
- the near impossibility of finding every single .fish file on the system
- the performance impact

> This could also be used to look at aliases, since I usually forget them when I am trying to learn them.

Sounds like a slightly rare but interesting case.

> do you have fd installed on your Mac?

Yes, if you didn't know, fzf.fish requires `fd` for the find file functionality!

> however I don't know how this would work on windows

Why wouldn't it work on Windows?

Overall I am a bit hesitant about adding this feature due to how rarely people will use it and how difficult it is to implement. But let me think about it more and get back to you! I really appreciate your enthusiasm and am sorry that I am telling you this now after you have already opened a PR.

matthiasdebernardini:
Hi, no problem I enjoyed working on this for myself. 

> the near impossibility of finding every single .fish file on the system

[...]

PR #29, opened 2020-09-12
print and search abbr list feature 

matthiasdebernardini:
[implementing this request](https://github.com/patrickf3139/fzf.fish/issues/28)

I am pretty new with fish and its plugins but I think I was able to implement the abbr feature (somewhat).

 I only got it to work on my system as it makes some assumptions which are;
1) the location of your abbreviations file (and that there is only one file with abbrs) and that its name matches abbr somehow
2) unsure if I am using fd-find interoperably, mine is the rust version
3) no way to configure this (maybe the user can configure their own path and if their favourite cli tool) 

I would however like some help polishing this up since I've hit some rough edges


patrickf3139:
Ah oh no, I should have responded sooner to you on #28 before you put in all this time. Let's discuss the feature's viability at a high level in the issue before we work out the code.

conf.d/fzf.fish
```diff
@@ -4,6 +4,7 @@ if not set --query fzf_fish_custom_keybindings
     bind \cf '__fzf_search_current_dir'
     bind \cr '__fzf_search_history'
     bind \cv '__fzf_search_shell_variables'
+    bind \cb '__fzf_search_abbreviations'
```
patrickf3139, line 7: I love how this keybinding is mnemonically inspired: b for 'bbreviations. However, it is found to move left. I wonder if people would miss using Ctrl+b to move left? Just a thought--not asking you to change it yet.

PR #30, opened 2020-09-22
Lots of changes

IlanCosman:
I have organized the changes into what I see as controversial versus uncontroversial ones. First comes changes to many files, then for individual files.

## Uncontroversial

1. Use the builtin `command --search` instead of `which`. Much faster and removes dependencies.

### `README`

1. Use prettier to format the README. Ensures consistent style, spacing, etc.

2. Reduce html use.

3. Use — (actual emdash character) instead of &mdash\;

### `_fzf_preview_file`

1. Use the builtin `test` many times instead `command ls -o -d "$file_path" | string sub --length 1`. Running `test` many times is much faster and less hacky. I have tested this by creating a named pipe with `mknod named-pipe-identifier p` and then spamming `time _fzf_preview_file named-pipe-identifier`. 200 microseconds vs. 3ms on my arch vm.

### `_fzf_report_file_type`

1. Use `--no-scope-shadowing` so that one does not need to pass `"$file_path"` every time in `_fzf_preview_file`.

## Controversial

1. Remove second dunderscore from the beginning of function names. Different fish plugins disagree on this, but in my opinion a second dunderscore is unnecessary, and is in general used by fish itself for it's functions, creating unnecessary overlap.

2. Use short flags for hyper-common options for hyper-common commands like `set` and `function`. I agree in general that long form flags are better, but if one is fish scripting one should know the common `set` and `function` flags down cold. Using `set --local` and `function --description` every time just increases the line length.

3. Related to the last point, use `set -l` when one means `set -l`. `set` and `set -l` do not always have the same meaning and clarity is good.

4. Also put function arguments before description. Arguments are more important and shorter for the person reading the code.

5. `camelCase` local variables, `snake_case` global and universal ones. This increases specification and clarity.

6. Use

   ```fish
   if set -l foo (bar)
       do stuff
   end
   ```

   instead of

   ```fish
   set -l foo (bar)
   if test $status -eq 0
       do stuff
   end
   ```

   This is a tiny bit faster and removes extra code.

7. Related to the last point, in multiple places you check for git repo like so:

   ```fish
   if not git rev-parse --git-dir >/dev/null 2>&1
       error
   else
       git do stuff | fzf
   end
   ```

   Instead do

   ```fish
   if not set -l gitDoStuff (git do stuff 2>/dev/null)
       error
   else
       printf '%s\n' $gitDoStuff | fzf
   end
   ```

   This is much faster because running any git command checks for a git repo in the same amount of time. Thus you're saving on running a git command twice if it turns out you are in a git repo.

### `LICENSE`

1. Convert license to markdown. As with `README`, `CONTRIBUTING`, `CHANGELOG` and the rest of the family, `LICENSE` is meant to be read by humans (granted no one ever actually reads the MIT license because we all know what it says but still). Anything that is meant to be read by humans should be rendered by the browser into good looking html, not displayed as plain text.

### `README`

1. Replace traditional links with [reference links](https://daringfireball.net/projects/markdown/syntax#link). There are a few links that are used many times (conf.d/fzf.fish, junegunn/fzf, jethrokuan/fzf etc). Reference links save you from having to put the link next to them every single time. Just put it at the bottom and you're good to go. I have chosen to also use reference format for non-reused links because reference links are vastly more [WYSIWYG](https://en.wikipedia.org/wiki/WYSIWYG).

---

I believe those are all the changes I made. If some of these changes seem annoyingly nitpicky or from an individual's style, feel free to take or leave any of them. I hope this PR doesn't come across as barging in and enforcing my style; It's all in good faith. I really appreciate your work for the fish community and the work you've put into this plugin :smile:


patrickf3139:
@IlanCosman you have left me speechless. 🤯  To say I'm very very grateful for this huge, high quality, thoroughly documented PR is an understatement. So instead, I'll try it with caps: THANK YOU! Thank you for making this plugin better, and more importantly, thank you for educating me on the very best of fish coding practices. 🙏

I've already taken a brief look and I agree with at least 90% of the changes. The last 10% I'll comment on and we can discuss what to do.

LICENSE.md
```diff
@@ -1,6 +1,6 @@
-MIT License
+# MIT License
```
patrickf3139, line 1: Love it! Looks way better in the browser now :)

README.md
```diff
@@ -1,101 +1,147 @@
-<h1 align="center">
-    fzf.fish
-</h1>
+<div align="center">
 
-A plugin that integrates [fzf](https://github.com/junegunn/fzf) into your [fish](http://fishshell.com) workflow. Comes with handy functions&mdash;each with its own mnemonic key binding&mdash;that use fzf to
```
patrickf3139, line 5: ah the reason I used &mdash is because it's a different character than just -. - is a bit too short and I think people are more used to seeing the longer dash. What do you think?

functions/__fzf_display_value_or_error.fish
```diff
@@ -1,10 +0,0 @@
-# helper function for __fzf_search_shell_variables
```
patrickf3139, line 1: Lol I think you basically touched 90% of my original code 😛 
Not a problem but I find it kinda funny.

functions/_fzf_search_shell_variables.fish
```diff
@@ -1,20 +1,15 @@
-function __fzf_search_shell_variables --description "Search and inspect shell variables using fzf. Insert the selected variable into the commandline at the cursor."
+function _fzf_search_shell_variables -d "Search and inspect shell variables using fzf. Insert the selected variable into the commandline at the cursor."
```
patrickf3139, line 1: Why switch to the short option? My rule of thumb is to always use the long form of options when scripting and use short hand when using the shell interactively.

functions/_fzf_preview_file.fish
```diff
@@ -0,0 +1,29 @@
+# helper function for _fzf_search_current_dir
+function _fzf_preview_file -a filePath -d "Prints a preview for the given file based on its file type."
+    if test -f "$filePath" # regular file
```
patrickf3139, line 3: Oh wow can't believe I didn't know `test` could tell you file types! Amazing!

README.md
IlanCosman: If you take a look at my README versus yours (the rendered versions), the emdashes look the same. In many text editors however they appear almost exactly the same size.

functions/_fzf_report_file_type.fish
```diff
@@ -0,0 +1,6 @@
+# helper function for _fzf_preview_file
+function _fzf_report_file_type -a fileType --no-scope-shadowing -d "Explain the file type for a file."
```
patrickf3139, line 2: --no-scope-shadowing is very clever 😀

functions/_fzf_search_shell_variables.fish
IlanCosman: You're not wrong. I'd agree that long options _are_ better for scripting in general. But I personally make an exception for the the most insanely common options like `set -l` or `functions -d`. You may feel differently and choose to be 100% consistent, but I think it's better to just save the line length since everyone contributing _should_ know these ultra basic options. To me it's like `rm -r`.

IlanCosman:
Also, as a shameless plug: I notice you're using the vastly inferior bobthefish in your screenshots. Have you considered using the the fish prompt of the gods, [Tide](https://github.com/IlanCosman/tide)? 😂

README.md
patrickf3139: Oh I didn't realize that! Thanks

patrickf3139:
> Have you considered using the the fish prompt of the gods, Tide? 😂

Oh Ilan, after this, I owe you a huge favor! I will not only switch to tide (it's already on my todo list to try, just never got around to it), but I'll update every single screenshot to use it and even make a brief mention!

IlanCosman:
Currently we display the date and time that a command was run in the `search_history` function. The default fzf bindings don't do this, nor does jethrokuan's, nor do the common bindings for other shells asfaik. I believe we should remove this feature as I don't think it's very useful and it complicates the code slightly by necessitating the next line to parse the output of fzf.

functions/_fzf_search_history.fish
```diff
@@ -0,0 +1,13 @@
+# originally implemented and transposed from https://github.com/patrickf3139/dotfiles/pull/11
+function _fzf_search_history -d "Search command history using fzf. Replace the commandline with the selected command."
+    # history merge incorporates history changes from other fish sessions
+    history merge
+
+    # Reference https://devhints.io/strftime to understand strftime format symbols
+    if set -l commandWithTs (history --show-time="%m/%e %H:%M:%S | " | fzf --tiebreak=index --query=(commandline))
```
patrickf3139, line 7: Are you sure you don't need `--null`? I'm concerned that if you have multi-line history fzf will get confused and break up the command.

IlanCosman: Oh really good point, I overlooked that. I'll test it out real quick.

patrickf3139:
Hey Ilan, let's take a step back. Most things look good and I appreciate all the hard work. However, I don't want to merge this massive PR. I know you already put a ton of work into it and it is thoroughly documented in the first post but I'd prefer if we broke the many changes down into bite sized PRs with a clear purpose.
Reasons
- easier for us to review and test => less bugs
- clearer git history and documentation
- easier for us to communicate back and forth about the changes

What do you think? If you agree, I think we can start by merging the uncontroversial changes then the controversial changes in that order, starting with a PR on

> Use the builtin command --search instead of which. Much faster and removes dependencies.

IlanCosman:
Yah that's entirely reasonable. I originally intended to change only small pieces but then things got out of hand 😂 I'll create some new PRs

patrickf3139:
Yeah I know that feeling.
Cool! Thanks for being so reasonable to work with! Can't wait to start merging :)

patrickf3139:
I'll keep referring back to this PR as a consolidated list of potential changes. And we can also discuss here whether or not we want a certain change before you open a PR so I don't want to cruelly reject a PR you put lots of hard work into and make you disappointed.

For now, here are the changes I will for sure approve to get you started:
1. Use the builtin command --search instead of which. Much faster and removes dependencies.
2. Most if not all the changes to the README
3. Everything you changed for `_fzf_preview_file`
4. Remove the second underscore from the beginning of function names.
5. Put function arguments before description.

patrickf3139:
Dang you're fast! 1 more I'll definitely approve:
Using

```
if set -l foo (bar)
    do stuff
end
```
instead of
```
set -l foo (bar)
if test $status -eq 0
    do stuff
end
```

Btw, for the dunderscore rename, I wonder if it's possible to keep git history when renaming? Currently researching. S.O. says it's not really possible but I once found a blog post where the guy said it was and I was able to reproduce his steps. Can't find it anymore 😢

daanzu:
> Currently we display the date and time that a command was run in the `search_history` function. The default fzf bindings don't do this, nor does jethrokuan's, nor do the common bindings for other shells asfaik. I believe we should remove this feature as I don't think it's very useful and it complicates the code slightly by necessitating the next line to parse the output of fzf.

FWIW, I think the date and time stamp can be helpful. I use the command search history frequently to find long command lines I've used in the past, and due to the fuzzy finding, they can be out of time order, so it can be helpful to use the time to find the right one (the one from yesterday not last week).

IlanCosman:
That's not a bad point. I still don't think it adds much value however. How many of those long line commands are you going to have? You'll probably be able to see the correct one near to top. And if you don't you just type a few more letters and it's there. I'm not vigorously opposed to having a timestamp though 🤷‍♂️

patrickf3139:
Oh hi there David, thanks for being a second pair of eyes!
Yeah I want to keep that functionality for
1) preservation of order of commands
2) being able to filter by date, e.g. if I want to look through only the commands I ran this past Monday

patrickf3139:
Hey @IlanCosman , thanks again for all your help. As we discussed, I looked into the rest of your proposed changes. It's a good thing you stopped at that last one because I've decided to hold off on all of them except for putting function arguments before description (btw, I'd say that one is uncontroversial as I did a survey of over 10 popular repos and 9/10 did it this way).

I at least owe you some explanation since you already put a lot of work and thought into it, so here it goes.

> `_fzf_report_file_type`: Use `--no-scope-shadowing` so that one does not need to pass "$file_path" every time in _fzf_preview_file.

I think functions should rarely be looking outside their internal function scope to find arguments that were meant to be passed in. This allows the caller of the function to rename its arguments and the function itself to rename its parameters, preventing brittleness. It also makes the function code a lot easier to read, as otherwise the reader would have to hop between two files to understand the code. I would consider using `--no-scope-shadowing` if `_fzf_report_file_type` was nested in another function.

> Remove second dunderscore from the beginning of function names

I did a survey of all the popular fzf plugins and they all use double underscores. Yes, it kind of conflicts with internal fish functions but I did also prefix the function names with fzf so there zero chance of namespace conflict.

> Use short flags for hyper-common options for hyper-common commands like set and function

I have to assume that there's going to be someone newish to fish reading my code and won't know the short options. Heck, that could be me in 3 years!

> Related to the last point, use set -l when one means set -l. set and set -l

I'd consider this if I had a function over 100 lines, but for shorter functions I actually do prefer to use `set` precisely because I do want all my internal function variables to be accessible throughout that function.

> camelCase local variables, snake_case global and universal ones

Never seen anyone else do that. And it's confusing to have to switch back and forth between them when writing the code.

PR #34, opened 2020-09-22
Format readme with prettier, use reference links

IlanCosman:
1. Use prettier to format the README. Ensures consistent style, spacing, etc.
2. Reduce HTML use.
3. Use — (actual emdash character) instead of `&mdash;`.

Broken out from #30.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
-<h1 align="center">
-    fzf.fish
-</h1>
+<div align="center">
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,101 +1,147 @@
-<h1 align="center">
-    fzf.fish
-</h1>
+<div align="center">
 
-A plugin that integrates [fzf](https://github.com/junegunn/fzf) into your [fish](http://fishshell.com) workflow. Comes with handy functions&mdash;each with its own mnemonic key binding&mdash;that use fzf to
```
patrickf3139, line 5: I've confirmed that - renders to the actual emdash even though in the text editor it looks like a short dash.

PR #39, opened 2020-09-26
put $fish_key_bindings inside quotes

uiae:
This commit fixes startup error messages of fzf.fish in case
$fish_key_bindings is not set.
If $fish_key_bindings is not set to a value, fzf.fish would generate the
following error messages on startup:
[: Missing argument at index 2

~/.config/fish/conf.d/fzf.fish (line 12):
    if [ $fish_key_bindings = 'fish_vi_key_bindings' ]
       ^
from sourcing file ~/.config/fish/conf.d/fzf.fish
	called on line 294 of file /usr/share/fish/config.fish
from sourcing file /usr/share/fish/config.fish
	called during startup

### DIFF
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -9,7 +9,7 @@ if not set --query fzf_fish_custom_keybindings
     bind \e\cs '__fzf_search_git_status'
 
     # set up the same key bindings for insert mode if using fish_vi_key_bindings
-    if [ $fish_key_bindings = 'fish_vi_key_bindings' ]
+    if [ "$fish_key_bindings" = 'fish_vi_key_bindings' ]
         bind --mode insert \cf '__fzf_search_current_dir'
         bind --mode insert \cr '__fzf_search_history'
         bind --mode insert \cv '__fzf_search_shell_variables'


patrickf3139:
Hi @ uiae, change looks good, but I'm curious--why does your fish environment not have a `fish_key_bindings` variable set?

uiae:
Hi patrickf3139, thank you for merging the PR!

You are right, $fish_key_bindings should be set in normal setups of fish.

I stumbled upon this incident while running fish for the first time on a new machine. I migrated all plugins (including fzf) and config.fish to the new machine but not ~/.config/fish/fish_variables.

$fish_key_bindings is set inside ~/.config/fish/fish_variables. If this file does not exist, fish will create it and set $fish_key_bindings to 'fish_default_key_bindings'. Fish will do this _after_ fzf.fish is loaded. Therefore, fzf.fish is loaded before $fish_key_bindings is set, which generates the above-mentioned error message. All consecutive startups of fish shell and fzf.fish are be fine as ~/.config/fish/fish_variables now exists and $fish_key_bindings is set.

So steps to reproduce are:
 - delete ~/.config/fish/fish_variables
 - close all running fish shells
 - open a fish shell

patrickf3139:
Ah that makes sense. Good fix! Even if no one ever deletes that file, I think it makes sense to quote the variable anyway! Thank you uiae.

PR #40, opened 2020-09-29
Always repaint in search_git_status

IlanCosman:
I mentioned this earlier and forgot. When running the `search_git_status` function, repaint should always be used (like in `search_git_log`) because something will always be printed.

Broken out from #30


patrickf3139:
Thank you so much for another bug fix!

PR #45, opened 2020-10-05
Stylistic and more idiomatic code

IlanCosman:
A few incredibly minor edits:

1. Use `test` instead of `[ ... ]`, in accordance with documentation and the rest of the code. https://github.com/fish-shell/fish-shell/issues/1850 says that use of the square brackets for test is highly discouraged.
2. Use `2>/dev/null` instead of `2> /dev/null`. This is preferred by `fish_indent`. I can demonstrate this if you wish.
3. Use `string escape "$blah"` instead of `echo $blah | string escape`


patrickf3139:
Hi again, boy how many times have you read through my code? Really appreciate my thoroughness.

PR #52, opened 2020-10-20
Don't suggest reporting an issue for unknown file types in __fzf_preview_file

IlanCosman:
This was some remnant code we failed to remove when switching from `ls` to `test`. I have chosen to not direct people to open an issue with this repository because there isn't really anything we could do--we are using all available file `test`s already.


patrickf3139:
Good point--there's no reason to have them open an issue that is almost impossible for us to fix.

PR #58, opened 2020-11-17
Use fzf_uninstall event to erase bindings

jorgebucaran:
Using the `fzf_uninstall` event is the best way to erase key bindings and cleanup resources. 

Fisher no longer supports `uninstall.fish`, so that file was removed as well. If we don't remove this file, it will be copied to `$__fish_config_dir`, potentially overwriting any existing `uninstall.fish` file in there.


patrickf3139 (approved):
Thank you Jorge!

jorgebucaran:
My pleasure! 💯

PR #59, opened 2020-11-17
Only support Fisher

patrickf3139:


bucaran:
This is a shame, after you opened https://github.com/oh-my-fish/oh-my-fish/issues/788 I thought you were honestly interested on the community's wellbeing. I guess we were all wrong.

The good part is that it only took a couple weeks before we could see your real intention.

PatrickF1:
^ Removed unnecessarily harsh and demeaning comment from bucaran.

PR #62, opened 2020-11-20
Set the shell to fish in __fzf_search_git_log

benjamineskola:
This was causing a bash syntax error as it was failing to interpret the fish syntax used for subcommands on line 9 (`git show` etc).

### DIFF
--- a/functions/__fzf_search_git_log.fish
+++ b/functions/__fzf_search_git_log.fish
@@ -2,6 +2,10 @@ function __fzf_search_git_log --description "Search the git log of the current g
     if not git rev-parse --git-dir >/dev/null 2>&1
         echo '__fzf_search_git_log: Not in a git repository.' >&2
     else
+        # Make sure that fzf uses fish to execute git show.
+        # See similar comment in __fzf_search_shell_variables.fish.
+        set --local --export SHELL (command --search fish)
+
         set selected_log_line (
             # see documentation for git format placeholders at https://git-scm.com/docs/git-log#Documentation/git-log.txt-emnem
             # %h gives you the abbreviated commit hash, which is useful for saving screen space, but we will have to expand it later below


functions/__fzf_search_git_log.fish
```diff
@@ -1,4 +1,8 @@
 function __fzf_search_git_log --description "Search the git log of the current git repository. Insert the selected commit hash into the commandline at the cursor."
+    # Make sure that fzf uses fish to execute git show.
```
PatrickF1, line 2: Could you move this logic into the else statement, please?

PatrickF1 (approved):
Hi Benjamin, thanks so much for finding and fixing this bug fix! Nice work :)

IlanCosman:
It would be good if we had a more global way of doing this for all fzf.fish invocations of fzf. We have like 5 instances of `set --local --export SHELL (command --search fish)`.

PatrickF1:
@IlanCosman I had the same thought. I can't just set it in conf/fzf.fish because that can cause problems for people whose main shell isn't fish. I suppose I could make a helper called `__fzf_force_use_fish` and call it as necessary. That'd be somewhat more consolidated. What do you think?

IlanCosman:
I don't know that there's actually a nice way of doing this unfortunately. Your method just shortens the line length (unless I'm misunderstanding), but it would be nice if we could just not say it at all in the `_search` functions.

A solution could be having a functions to which one passes the name of the actual function you want called. Better explanation:

Instead of 
```fish
# etc.
```

we do
```fish
bind \cf '__fzf current_dir'
bind \cr '__fzf history'
```

then `__fzf` would look like
```fish
function __fzf -a name
    __fzf_search_$name
end
```

Essentially it prepends the command to every search function. This might increase latency by a minuscule amount. Maybe the status quo is just fine though.

PatrickF1:
How clever of you! Yes that would work. I don't mind the increased latency but it does add another layer of complexity to the code, however thin. Which makes it a bit harder for everyone to maintain. I'll sleep on it. Thanks Ilan!

PR #65, opened 2020-12-07
Ensure history uses fish builtin

reynn:
- specify builtin for history to ensure no aliases

This helps prevent issues with ctrl+r if `history` is aliased to something else.


functions/__fzf_search_history.fish
```diff
@@ -1,9 +1,9 @@
 function __fzf_search_history --description "Search command history using fzf. Replace the commandline with the selected command."
     # history merge incorporates history changes from other fish sessions
-    history merge
+    builtin history merge
```
PatrickF1, line 3: Are you sure we don't want `command` instead of `builtin`. Here, we are using the fish provided version of history and command forces us to use that.

PatrickF1: My bad--sorry--you are right. History is a builtin and so we want `builtin history`. Thanks for your PR!

PatrickF1 (approved):

functions/__fzf_search_history.fish
reynn: Thanks! Here is a reference to the builtin docs in case anyone else comes upon this https://fishshell.com/docs/current/cmds/builtin.html

PR #71, opened 2020-12-31
Show more info when previewing variables and fix missing variables or wrong values

kidonng:
set --show shows more variable info about a variable:

- Scope (global or universal)
- Exported or unexported
- Element count
- Show one element per line, more easy to read

image

image


functions/__fzf_display_value_or_error.fish
```diff
@@ -1,7 +1,7 @@
 # helper function for __fzf_search_shell_variables
 function __fzf_display_value_or_error --argument-names variable_name --description "Displays either the value of the variable passed in, or an informative message if it is not available."
     if set --query $variable_name
-        echo $$variable_name
+        set -S $variable_name | string replace -r "^\\\$$variable_name(: )?" ""
```
PatrickF1, line 4: Can you use `--show` instead of -S, please?

PatrickF1:
Testing this out now, thanks for the awesome idea @kidonng!

functions/__fzf_display_value_or_error.fish
PatrickF1: What's the point of the call to string replace?

kidonng: > Can you use `--show` instead of -S, please?

Sure thing!

> What's the point of the call to string replace?

Make the output more concise. This is the unmodified output of `set --show`:

PatrickF1: Nice, that's very clever, much easier to read indeed

PatrickF1:
So...here's the thing. I'm conflicted. You've put in tons of work but there's a serious problem with using `set --show`. It's not helpful to view the variable scopes because only exported variables show up due to the nature of the `set` command being executed by fzf.
The good news is that using `set --show` is still valuable because it makes array variables much easier to read.

If we can cut out the variable scope info and keep the array variable broken up by line then this'll be a great addition. What do you think?

kidonng:
@PatrickF1 I think we can just dump current variables using `psub`. This gives almost all current variables' info (except for `_` which is always `fish` and `history` which is isn't in `set --show`'s output). For example, we now have correct `SHLVL`:

image

PatrickF1:
Wow that's pretty clever if not a bit too complicated and hacky for my tastes. Let's just not filter out the local variables.
To be honest, I think `set --show`'s output contains too much unnecessary information most of the time and is hard to parse. Most users, I bet, just want to know the value of the variable. With that in mind, I think there's still something to salvage here, and that's to better delineate the values of arrays.

Do you think you can change your code to do just that? Print out each value in the variable on a separate line? Sorry for scrapping the hard work you've put in already. Hopefully we've both learned from it and it's sharpened our shell skills!

kidonng:
> Let's just not filter out the local variables

I'm not. There's no way to show local variables because we are in a function.

> Most users, I bet, just want to know the value of the variable. Scope is generally irrelevant when you're just inspecting the values in an interactive shell.

The scope and element count only occupy the first line. And the index is useful in case you want to slice (`[2..5]`) or remove specific element (`set -e variable[3]`). I can't see why users will find this unhelpful.

> Furthermore, for variables that exist in multiple scopes, it's printed multiple times and that's confusing.

Since we don't show local variables, this is very rare. And AFAIK Fish is going to change this behavior. Can you provide some examples when setting both a global and universal variable is necessary?

> set --show's output contains too much unnecessary information... However, I think there's still something to salvage here, and that's to better delineate the values of arrays.

So I just stripped colon and veritcal line (`|`), how does it look now?

image

kidonng:
I think we can do better. I will push some updates later.

PatrickF1:
Let's back up and make sure we're designing this for what the typical user of this feature wants. I picture they are in their shell and want to inspect the current variables. They'd want and expect to be able to find all the variables that they can currently access when they hit ctrl+v (or whatever they used for their keybinding) and see quickly read their value. i.e. I'd write the user story as "As a fish user, I want to be able to quickly find and inspect any of the shell variables in my current environment". So, it'd be mighty confusing (not to mention backwards incompatible) for us to sudden start removing local variables from fzf, even if we can't give them a value. And that's why I wrote up that special function `__fzf_display_value_or_error.fish` just to tell the user the variable they want can't be inspected because it's local.

PatrickF1:
Hey @kidonng how are you feeling? If this is getting frustrating for you, I'm sorry, but I have to be very picky to protect the plugin in the best interest of ALL potential and current users. If you're tired of going back and forth, let me know and I'll just modify your PR and merge it with your name on it so you still get credit.

kidonng:
> Let's back up and make sure we're designing this for what the typical user of this feature wants. I picture they are in their shell and want to inspect the current variables. They'd want and expect to be able to find all the variables that they can currently access when they hit ctrl+v (or whatever they used for their keybinding) and see quickly read their value. i.e. I'd write the user story as "As a fish user, I want to be able to quickly find and inspect any of the shell variables in my current environment". So, it'd be mighty confusing (not to mention backwards incompatible) for us to sudden start removing local variables from fzf, even if we can't give them a value. And that's why I wrote up that special function `__fzf_display_value_or_error.fish` just to tell the user the variable they want to see can't be inspected because it's local.

Thanks for explaining so thoroughly, but you missed a point. `set --names` doesn't give **any** local variables **outside of the function** from the very beginning (it only shows `$argv` and `$SHELL`, the latter is being removed in this PR). And since fzf.fish doesn't dump current environment, the user is **not inspecting the current variables**, but only variables available from a forked Fish shell. Now this PR is more than just `set --show`, it acutally **fixes the bug**.

`__fzf_display_value_or_error` actually only works for **unexported** variables (because forked shell can't see them), not **local** variables (because there are none).

> Set always shows at least three lines for scope: 1 for local, 1 for global, 1 for universal. That's 3 lines too many. You're right about the index. Let's keep the index.

There are **no** local variables except `$argv`.

> See history, status, pipestatus. I'm sure there's more.

`$history` is a intersting one. It doesn't appear in `set --show`, we have to explicity use `set --show history`. But I just did some experiment and find that any attempt to workaround this will result in chaing `$status` and `$pipestatus`. We can display a error message for it, just like `__fzf_display_value_or_error`.

And `set --show | psub` seems like the only way to retain `$status` and `$pipestatus`:

image

image

And they are not good examples of "setting both a global and universal variable is necessary". Let me repeat myself: there are **no** local variables except `$argv`. These will never be universal variables either. And even you does provide real examples, they will actually consolidate the need to show scopes.

image

> Also, some users (like me actually back when I first learning fish) might have a bunch of set -xg in their config.fish.

We **dumped the variables**, `config.fish` or `conf.d` scripts won't affect anything.

> > Turns out the only way to do this is through set --show. We can't just parse $VARIABLE because we have no idea how they are separated, only Fish knows.
> 
> https://stackoverflow.com/questions/49223562/fish-iterate-over-a-string

I know how to iterate. But in order to display nicely it will need more efforts, and there are some edge cases we need to handle. And we lost `$status` and `$pipestatus`, probably more.

> > Now that colon and veritcal line (|) are also stripped (besides variable name), how does it look? IMHO this is the best compromise to be made.
> 
> Nice work! That's beautiful and I think that's very close to what I want. I still want to remove the lines printing scope though, so I think we need to print out the variable ourselves by following the SO post code I linked above.

If you **really** think that informational line is disturbing, that's fine. As for me, this PR is mostly done.

kidonng:
Summarize:

- No local variables can be shown. Unless you build a custom Fish shell.
- Some variables may still be missing (notably `$history`), we may want to show a message for this.
- The informational line will be kept.
  If there are both a global variable and a universal variable of the same name, most of the time it means users messed up their environment and this serves as a handy tip to remind them to fix it.
  At other time, it will only occupy the first line, but provides three piece of valuable info at a glance. I just don't understand why would this be annoying.
- Use `set --show` instead of iterate/expand variables. This prevents mistakes and preserve useful variables such as `$status` and `$pipestatus`.

I don't want to be stubborn, after all this is your project, not mine. But I really need a convincing reason why something is bad or can be better. And the currect state is way better than before this PR, where there is only a vairable from a forked shell being shown.

PatrickF1:
> set --names doesn't give any local variables outside of the function from the very beginning 

Oh wow you're right. This whole time I was testing it by doing
```
set abcdefg
# then hit Ctrl+v and search for abcdefg and find it
```
but I just realized that it's being set in global not local. Ok I think we haven't been on the same page at all due to my misunderstanding. Let me re-read everything and move on from here. Sorry about my mistake!

> And since fzf.fish doesn't dump current environment, the user is not inspecting the current variables, but only variables available from a forked Fish shell. Now this PR is more than just set --show, it acutally fixes the bug.

You are right. We need to keep psub because for important variables such as status and pipestatus, we need it!

functions/__fzf_search_shell_variables.fish
```diff
@@ -1,16 +1,14 @@
...
     # Pipe the names of all shell variables to fzf and attempt to display the value
     # of the selected variable in fzf's preview window.
-    # Non-exported variables will not be accessible to the fzf process, in which case
-    # __echo_value_or_print_message will print an informative message in lieu of the value.
+    # Using string match to filter variables.
+    # Using string repalace to simplify output.
     set variable_name (
         set --names |
-        fzf --preview '__fzf_display_value_or_error {}'
+        fzf --preview "cat "(set --show | psub)" |
```
PatrickF1, line 8: Do you think it's possible to reduce calls to `set --show | psub` by calling it above and reusing the file returned by psub?

kidonng:
```
❯ type psub
psub is a function with definition
# Defined in /usr/local/Cellar/fish/3.1.2/share/fish/functions/psub.fish @ line 1
function psub --description 'Read from stdin into a file and output the filename. Remove the file when the command that called psub exits.'
```

PatrickF1: Darn...just to confirm I also tested it :(
```
~ ❯ set file_name (echo hi | psub)
~ ❯ cat $file_name
cat: /var/folders/kv/6ynq38897hl_w48jfgzl8w080000gp/T//.psub.S7Je5fqOCy: No such file or directory
```

PatrickF1:
Btw can you clarify what edge cases there are?

>  But in order to display nicely it will need more efforts, and there are some edge cases we need to handle.

kidonng:
Spaces, line breaks, empty lists, to name a few.

functions/__fzf_search_shell_variables.fish
kidonng: > reduce calls to `set --show | psub`

You misread the code. `set --show | psub` only excutes once.

functions/__fzf_display_value_or_error.fish
```diff
@@ -1,10 +1,12 @@
 # helper function for __fzf_search_shell_variables
-function __fzf_display_value_or_error --argument-names variable_name --description "Displays either the value of the variable passed in, or an informative message if it is not available."
-    if set --query $variable_name
-        echo $$variable_name
+function __fzf_display_value_or_error --argument-names variable_name variable_file --description "Displays either the value of the variable passed in, or an informative message if it is not available."
+    set matches (cat $variable_file | string match --regex "^\\\$$variable_name(?::|\[).+")
```
jorgebucaran, line 3: @kidonng [UUOC](https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat). This will do it:

```fish
string match --regex "^\\\$$variable_name(?::|\[).+" <$variable_file
```

jorgebucaran (changes requested):

functions/__fzf_display_value_or_error.fish
PatrickF1: Thanks for helping me review this, Jorge! Please also chime in on the feature design; I know you have a good eye for these things.

functions/__fzf_display_value_or_error.fish
```diff
@@ -1,10 +1,12 @@
 # helper function for __fzf_search_shell_variables
-function __fzf_display_value_or_error --argument-names variable_name --description "Displays either the value of the variable passed in, or an informative message if it is not available."
-    if set --query $variable_name
-        echo $$variable_name
+function __fzf_display_value_or_error --argument-names variable_name variable_file --description "Displays either the value of the variable passed in, or an informative message if it is not available."
```
PatrickF1, line 2: Thanks for bringing back the helper file. Much more readable this way!

functions/__fzf_display_value_or_error.fish
```diff
@@ -1,10 +1,12 @@
...
+    set matches (cat $variable_file | string match --regex "^\\\$$variable_name(?::|\[).+")
+
+    if test -n "$matches"
+        string replace --regex "^\\\$$variable_name(?:: (.+)|(\[.+\]): \|(.+)\|)" '\$1\$2 \$3' $matches
     else
         set_color red
-        echo "$variable_name was not exported to this process so its value cannot be displayed." >&2
+        echo "$variable_name was not dumped so its value cannot be displayed." >&2
```
PatrickF1, line 9: The error message is an abstraction leak. The user doesn't know that we use `set --show | psub` so they will be confused by why the error message is saying it wasn't dumped.

kidonng: We does *export the variable into* a temporary file, but not in the meaning of *export a variable to a subprocess*. I was thinking that keeping the old message may mislead the user. But I guess it's probably no difference.

PatrickF1:
Can you rebase this on master so we can test more confidently? Thanks!

functions/__fzf_search_shell_variables.fish
```diff
@@ -1,16 +1,13 @@
...
-    # Non-exported variables will not be accessible to the fzf process, in which case
-    # __echo_value_or_print_message will print an informative message in lieu of the value.
+    # Non-dumped variables will not be accessible to the fzf process, in which case
+    # __fzf_display_value_or_error will print an informative message in lieu of the value.
+    # Make sure that fzf uses fish to execute __fzf_display_value_or_error, which
+    # is an autoloaded fish function so doesn't exist in other shells.
     set variable_name (
         set --names |
-        fzf --preview '__fzf_display_value_or_error {}'
+        SHELL=$__fish_bin_dir/fish fzf --preview '__fzf_display_value_or_error {} '(set --show | psub)
```
PatrickF1, line 10: Btw, how come you had to move setting SHELL to fish from up there to down here?

kidonng: Was thinking this can possibly remove the local `$SHELL`. Turns out it doesn't so I will revert it.

PatrickF1:
Hey can you rebase since i merged the last PR which touched the `functions/__fzf_search_shell_variables.fish`?
Also hey crazy idea, but what if we made the call to psub at the very top like so:
`bind \cv '__fzf_search_shell_variables (set --show | psub)'?

kidonng:
> 3\. Since you've already put so much work into this, would you mind starring this repo?

Sure! I usually doesn't star a repo if I'm using it or it is well-known, but it doesn't matter.

PatrickF1:
> Sure! I usually don't star a repo if I'm already using it or it is well-known, but it doesn't matter.
Thanks! That makes it all the more special to receive your star! FYI I'm asking for stars is to surpass https://github.com/jethrokuan/fzf. Jethro and I agree that this plugin should be the new de facto plugin in the fzf-fish world.

kidonng:
> 2\. crazy idea, but what if we made the call to psub at the very top like so: `bind \cv '__fzf_search_shell_variables (set --show | psub)`? This way, we can capture even local variables

It works! Should definitely give it a try.

conf.d/fzf.fish
```diff
@@ -3,7 +3,7 @@ if not set --query fzf_fish_custom_keybindings
     # \cf is Ctrl+f
     bind \cf '__fzf_search_current_dir'
     bind \cr '__fzf_search_history'
-    bind \cv '__fzf_search_shell_variables'
+    bind \cv '__fzf_search_shell_variables (set --show (set --names) | psub) (set --names)'
```
PatrickF1, line 6: You can probably just save that into a variable and then re-use it below.

kidonng: Then we will have an extra variable shown. And it doesn't save much.

```
> time set --names
________________________________________________________
Executed in  645.00 micros    fish           external
   usr time  394.00 micros  394.00 micros    0.00 micros
   sys time  240.00 micros  240.00 micros    0.00 micros
```

functions/__fzf_search_shell_variables.fish
```diff
@@ -1,19 +1,19 @@
 function __fzf_search_shell_variables --description "Search and inspect shell variables using fzf. Insert the selected variable into the commandline at the cursor."
```
PatrickF1, line 1: Let's declare the arguments and also set up a check for when there are no variables (for people who customized their key bindings) and do something like `echo "The signature of __fzf_search_shell_variables was changed in https://github.com/PatrickF1/fzf.fish/pull/71. Please see the latest conf.d/fzf.fish and update your key bindings."`

conf.d/fzf.fish
PatrickF1: Does it lead to an extra variable? I don't care about saving time but not having two copies of the string `'__fzf_search_shell_variables (set --show (set --names) | psub) (set --names)'`

kidonng: > '__fzf_search_shell_variables (set --show (set --names) | psub) (set --names)'

Oh, this can be avoided. I'm taking about the binded function.

kidonng: As explained before, `set --show (set --names)` shows `$history` but `set --show` doesn't.

functions/__fzf_search_shell_variables.fish
kidonng: @PatrickF1 With the newest change, `__fzf_display_value_or_error` becomes unnecessary. Is it fine to remove it (again)?

conf.d/fzf.fish
```diff
@@ -1,6 +1,6 @@
 # Set up the default, mnemonic key bindings unless the user has chosen to customize them
 if not set --query fzf_fish_custom_keybindings
-    set -l search_shell_variables '__fzf_search_shell_variables (set --show (set --names) | psub) (set --names)'
+    set -l search_shell_variables '__fzf_search_shell_variables (set --show | psub) (set --names)'
```
kidonng, line 3: Hmm, I'm not sure about this change. This means we need to dispaly a error message, presumably just `$history`.

PatrickF1: Yeah I'll just filter out history.

Don't know if you saw my commit comment but here it is
> don't try to show history variable because it adds complexity and because this plugin has a function dedicated to showing history already anyway

functions/__fzf_search_shell_variables.fish
```diff
@@ -10,20 +10,21 @@ function __fzf_search_shell_variables --argument-names variable_file --descripti
...
-    # variable in fzf's preview window.
+    # Exclude the history variable name from being piped into fzf because it's not included in
+    # $variable_file (the output of set --show). It's also not worth showing anyway as
+    # __fzf_search_history is a much better way to examine history.
     # We use the current token to pre-populate fzf's query. If the current token begins
     # with a $, we remove it from the query so that it will better match the variable names
     # and we put it back later when replacing the current token with the user's selection.
     set variable_name (
-        string collect $argv[2..-1] |
+        string collect $argv[2..-1] | grep --invert-match history |
```
kidonng, line 22: Use `string` instead?

PatrickF1: oh good idea, yeah less deps

kidonng:
Is this good to merge? I have the idea to allow mutliple selection accroding to https://github.com/PatrickF1/fzf.fish/pull/73#discussion_r552143666, but let's move it to a new PR.

PatrickF1:
Hmm I keep running into `string replace: Regular expression substitute error: requested value is not set` from even before I made my changes (I reverted back to 3c35eb8 and still get the error). Anyway, it's time for me to go to bed...but do you know what went wrong?

functions/__fzf_search_shell_variables.fish
```diff
@@ -1,24 +1,33 @@
...
+    # $variable_file (the output of set --show). It's also not worth showing anyway as
+    # __fzf_search_history is a much better way to examine history.
     # We use the current token to pre-populate fzf's query. If the current token begins
     # with a $, we remove it from the query so that it will better match the variable names
     # and we put it back later when replacing the current token with the user's selection.
     set variable_name (
-        set --names |
-        fzf --preview '__fzf_display_value_or_error {}' \
+        string collect $argv[2..-1] | string match --invert history |
+        fzf --preview "string match   --regex '^\\\${}(?::|\[).+' <$variable_file |
```
PatrickF1, line 24: Need a bit of documentation on what these the string match and string replace commands are doing.

kidonng:
I didn't get that error. Which revision are you at?

conf.d/fzf.fish
```diff
@@ -1,9 +1,10 @@
 # Set up the default, mnemonic key bindings unless the user has chosen to customize them
 if not set --query fzf_fish_custom_keybindings
+    set -l search_shell_variables '__fzf_search_shell_variables (set --show | psub) (set --names)'
```
PatrickF1, line 3: Can we also send the second argument into psub? Am worried about hitting argument limits. And also, then we'd be able to give an argument name to all the variables.

PatrickF1:
I tried 3c35eb8.
Also, yeah we're very close. I think after I fix whatever bug is causing it to not work and leave some more documentation, we are good to go!

conf.d/fzf.fish
kidonng: > Am worried about hitting argument limits

image

It's not like someone will have lots of variables.

kidonng:
> Also can you allow me to push please?

- You are already pushing to my branch 😅

- This switch is on by default
 
  image

conf.d/fzf.fish
PatrickF1: Ah good to know I don't need to worry about it. Sending it in as one file and then giving that file an argument name is self-documenting, however, so I still strongly prefer it.

functions/__fzf_search_shell_variables.fish
PatrickF1: Sorry I know you wanted to get rid of that extra file but I think we need a helper for the preview command so that we can more easily document the logic.

PatrickF1:
> You are already pushing to my branch 😅

I've been doing it through the Github UI. When I try to push from my terminal it says no access :(

kidonng:
I turned it off and on again. Can you confirm the issue is resolved?

conf.d/fzf.fish
kidonng: That's fine.

kidonng:
> Hmm I keep running into `string replace: Regular expression substitute error: requested value is not set` from even before I made my changes (I reverted back to [3c35eb8](https://github.com/PatrickF1/fzf.fish/commit/3c35eb82fbc53f19bd801c119fea3024cff559a4) and still get the error). Anyway, it's time for me to go to bed...but do you know what went wrong?

I can reproduce it with Fish 3.1.2 now. I think this is a Fish bug because it doesn't happen on Fish's master branch.

kidonng:
> `set --show`'s output contains too much unnecessary information

Wow, I literally just realized, this is true in Fish 3.1.2. The whole time I was using the master branch of Fish which doesn't show the "not set in <scope> scope" message.

functions/__fzf_search_shell_variables.fish
kidonng: Sure, but let's give it a name.

kidonng:
The filter has been improved and documented, not only does it work in Fish 3.1.2, but it also strip the uncessary info that is already removed in unreleased latest Fish.

PatrickF1:
Wow nice work! We are super close to merging (finally, hallelujah!). What was the reason it doesn't work on 3.1.2? Also I'm even more optimistic about merging if fish on master branch is less verbose about variable scopes.

PatrickF1:
My latest commit:
1. removed comment about the newlines in the guard statement
2. renamed arguments to be very clear about what they are
3. refactored out logic to exclude history from list of variable name to simplify the call to fzf
4. added documentation on top about how to call this function

Sorry it's untested so you might run into errors, Xuann

kidonng:
> What was the reason it doesn't work on 3.1.2? 

https://github.com/PatrickF1/fzf.fish/pull/71#issuecomment-755942287

PatrickF1:
Oh oops I missed that. Got it. FYI I'm gonna delete this comment, your comment, and edit my original question to reduce the amount of back-and-forth on this page.

PatrickF1:
Ugh status is still messed up...I don't suppose there's anything we can do to fix it since we're calling psub as early as we can already. Do you know why the status is being set to 0 though?

kidonng:
It's because I flipped them in 129ae3dd101c9f6730daf4ce07c65e83bd3c9a41

PatrickF1:
Oh of course! What would I do without you?
I'm about ready to merge. Are you sure the only variable that is missing or messed up is history?

kidonng:
I have counted it for countless times (bad pun).

PatrickF1 (approved):
Woohoo! What a long journey. I think we're done! * highfive * @kidonng

PR #72, opened 2021-01-03
Prepend ./ to single directory selections, allowing for quick cd

kidonng:
This PR adds the following feature to File paths (<kbd>Ctrl</kbd> + <kbd>F</kbd>):

A dot and a slash (`./`) will be prepended automatically if:

- There is no input in the command line (the user triggers the function with a empty input)
- The user only made one selection
- The only selection is a directory

This makes it possible to quick enter the selected directory by pressing <kbd>Enter</kbd> again after making the selection. This is because [Fish treats it as implict `cd`](https://fishshell.com/docs/current/cmds/cd.html):

> Note that the shell will attempt to change directory without requiring cd if the name of a directory is provided (starting with ., / or ~, or ending with /).

<details>
<summary>
Screencast (Outdated)
</summary>

https://user-images.githubusercontent.com/44045911/103480545-e5b12580-4e0f-11eb-813a-13a2143a565f.mov
</details>

If the user changed their mind, they can do anything with the folder just like normal. ~~The trailing slash doesn't make much difference most of the time.~~

### DIFF
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -9,9 +9,8 @@ function __fzf_search_current_dir --description "Search the current directory us
 
     if test $status -eq 0
         # If this function was triggered with an empty commandline and the only thing selected is a directory, then
-        # prepend a dot and a slash to the folder. Because fish will attempt to cd implicitly if a directory name
-        # is provided with a dot at the beginning, this allows the user to hit Enter one more time to quickly cd
-        # in the selected directory.
+        # prepend ./ to the dir path. Because fish will attempt to cd implicitly if a directory name starting with a dot
+        # is provided, this allows the user to hit Enter one more time to quickly cd into the selected directory.
         if test (count (commandline -o)) = 0 && test (count $file_paths_selected) = 1 && test -d $file_paths_selected
             set file_paths_selected ./$file_paths_selected
         end
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -8,6 +8,10 @@ function __fzf_search_current_dir --description "Search the current directory us
     )
 
     if test $status -eq 0
+        if test (count (commandline -poc)) = 0 && test (count $file_paths_selected) = 1 && test -d $file_paths_selected[1]
+            set file_paths_selected[1] $file_paths_selected[1]/
+        end
+
         for path in $file_paths_selected
             set escaped_path (string escape "$path")
             commandline --insert "$escaped_path "
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -12,7 +12,7 @@ function __fzf_search_current_dir --description "Search the current directory us
         # prepend a dot and a slash to the folder. Because fish will attempt to cd implicitly if a directory name
         # is provided with a dot at the beginning, this allows the user to hit Enter one more time to quickly cd
         # in the selected directory.
-        if __fish_is_token_n 1 && test (count $file_paths_selected) = 1 && test -d $file_paths_selected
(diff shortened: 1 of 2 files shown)


PatrickF1 (approved):
Oh man that is beautiful! Thank you @kidonng! Let me test real quick and then I'll merge.

jorgebucaran (approved):

PatrickF1:
I tested it on directories with spaces in the name not expecting it to work but it does...I don't understand why 😂

functions/__fzf_search_current_dir.fish
```diff
@@ -8,6 +8,13 @@ function __fzf_search_current_dir --description "Search the current directory us
     )
 
     if test $status -eq 0
+        # If this function was triggered with an empty commandline and the only thing selected is a directory, then
+        # append a slash to the folder. Because fish will attempt to cd implicitly if a directory name is provided with
+        # a / at the end, this allows the user to hit Enter one more time to quickly cd in the selected directory.
+        if __fish_is_token_n 1 && test (count $file_paths_selected) = 1 && test -d $file_paths_selected[1]
```
PatrickF1, line 14: Would the code be clearer if you removed [1] everywhere because the test (count $file_paths_selected) = 1 makes sure the variable is exactly one element long anyway?

PatrickF1, line 14: Is `__fish_is_token_n 1` really necessary?

kidonng:
> I tested it on directories with spaces in the name not expecting it to work but it does...I don't understand why 😂

The escaping is done below by the original code.

functions/__fzf_search_current_dir.fish
kidonng: Yes, this check for an empty commandline.

kidonng: Sure! I didn't notice that.

kidonng:
I changed my mind and decided to prepend a dot and a slash (`./`) instead. This prevents potential difference caused by trailing slash. And the users are more likely to have a better idea what this does for the first time.

PatrickF1:
> This prevents potential difference caused by trailing slash
Hmm I was actually starting to think about just appending /  for all directories selected. What potential problems does trailing / cause? And how is ./dir better?

functions/__fzf_search_current_dir.fish
```diff
@@ -8,6 +8,14 @@ function __fzf_search_current_dir --description "Search the current directory us
     )
 
     if test $status -eq 0
+        # If this function was triggered with an empty commandline and the only thing selected is a directory, then
+        # prepend a dot and a slash to the folder. Because fish will attempt to cd implicitly if a directory name
+        # is provided with a dot at the beginning, this allows the user to hit Enter one more time to quickly cd
+        # in the selected directory.
+        if __fish_is_token_n 1 && test (count $file_paths_selected) = 1 && test -d $file_paths_selected
```
PatrickF1, line 15: I realized through testing __fish_is_token_n doesn't ensure the commandline is empty. If there is input after the cursor this block still executes.

kidonng:
> What potential problems does trailing / cause?

I can only remember one. Suppose we have the following directory:

```
.
├── link -> source
└── source
   ├── file1
   ├── file2
   └── file3
```

And you want to remove the link.

- `rm -rf link`

   ```
   .
   └── source
      ├── file1
      ├── file2
      └── file3
   ```

- `rm -rf link/`

   ```
   .
   └── link -> source
   ```

> And how is ./dir better?

Because it doesn't have trailing slash, while still being recognized by Fish.

functions/__fzf_search_current_dir.fish
kidonng: Nice catch! Actually I should just use `commandline -c`.

PR #73, opened 2021-01-04
Always pre-populate fzf's search query with current token

kidonng:
Currect only "A previously run command" uses existing input as fzf's default query. I think this should apply to all functions.

For variable searching I did some extra work for leading `$`:

https://user-images.githubusercontent.com/44045911/103555474-4feac880-4eeb-11eb-918c-97878fa43180.mov

### DIFF
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -15,10 +15,7 @@ function __fzf_search_current_dir --description "Search the current directory us
             set file_paths_selected ./$file_paths_selected
         end
 
-        for path in $file_paths_selected
-            set escaped_path (string escape "$path")
-            commandline --current-token --replace "$escaped_path "
-        end
+        commandline --current-token --replace (string escape $file_paths_selected | string join ' ')
     end
 
     commandline --function repaint
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -4,13 +4,13 @@ function __fzf_search_current_dir --description "Search the current directory us
     set --local --export SHELL (command --search fish)
     set file_paths_selected (
         fd --hidden --color=always --exclude=.git 2>/dev/null |
-        fzf --multi --ansi --preview='__fzf_preview_file {}'
+        fzf --multi --ansi --preview='__fzf_preview_file {}' --query=(commandline --current-token)
     )
 
     if test $status -eq 0
         for path in $file_paths_selected
             set escaped_path (string escape "$path")
-            commandline --insert "$escaped_path "
+            commandline --current-token --replace "$escaped_path "
         end
     end
 
--- a/functions/__fzf_search_git_log.fish
+++ b/functions/__fzf_search_git_log.fish
@@ -10,12 +10,12 @@ function __fzf_search_git_log --description "Search the git log of the current g
             # see documentation for git format placeholders at https://git-scm.com/docs/git-log#Documentation/git-log.txt-emnem
             # %h gives you the abbreviated commit hash, which is useful for saving screen space, but we will have to expand it later below
             git log --color=always --format=format:'%C(bold blue)%h%C(reset) - %C(cyan)%as%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)' | \
-            fzf --ansi --tiebreak=index --preview='git show --color=always (string split --max 1 " " {})[1]'
+            fzf --ansi --tiebreak=index --preview='git show --color=always (string split --max 1 " " {})[1]' --query=(commandline --current-token)
         )
         if test $status -eq 0
             set abbreviated_commit_hash (string split --max 1 " " $selected_log_line)[1]
             set commit_hash (git rev-parse $abbreviated_commit_hash)
-            commandline --insert $commit_hash
+            commandline --current-token --replace $commit_hash
         end
     end
 
--- a/functions/__fzf_search_shell_variables.fish
+++ b/functions/__fzf_search_shell_variables.fish
@@ -10,11 +10,16 @@ function __fzf_search_shell_variables --description "Search and inspect shell va
     # __echo_value_or_print_message will print an informative message in lieu of the value.
     set variable_name (
         set --names |
-        fzf --preview '__fzf_display_value_or_error {}'
+        fzf --preview '__fzf_display_value_or_error {}' \
+            --query=(commandline --current-token | read token && string replace '$' '' $token)
     )
 
     if test $status -eq 0
-        commandline --insert $variable_name
+        if string match -q '$*' $token
+            commandline --current-token --replace \$$variable_name
+        else
+            commandline --current-token --replace $variable_name
+        end
     end
 
     commandline --function repaint
--- a/functions/__fzf_search_shell_variables.fish
+++ b/functions/__fzf_search_shell_variables.fish
@@ -8,8 +8,8 @@ function __fzf_search_shell_variables --description "Search and inspect shell va
     # of the selected variable in fzf's preview window.
     # Non-exported variables will not be accessible to the fzf process, in which case
     # __echo_value_or_print_message will print an informative message in lieu of the value.
-    # We use the current token to pre-populate fzf's search input. In case the user has typed
-    # a $, we remove it so that the current token will more likely match the variable names
+    # We use the current token to pre-populate fzf's query. If the current token begins
+    # with a $, we remove it from the query so that it will better match the variable names
     # and we put it back later when replacing the current token with the user's selection.
     set variable_name (
         set --names |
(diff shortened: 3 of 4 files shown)


functions/__fzf_search_shell_variables.fish
```diff
@@ -10,11 +10,16 @@ function __fzf_search_shell_variables --description "Search and inspect shell va
     # __echo_value_or_print_message will print an informative message in lieu of the value.
     set variable_name (
         set --names |
-        fzf --preview '__fzf_display_value_or_error {}'
+        fzf --preview '__fzf_display_value_or_error {}' \
+            --query=(commandline --current-token | read token && string replace '$' '' $token)
```
PatrickF1, line 14: Can' you just do 
```
--query=(commandline --current-token | string replace '$' '' $token)
```?

**Comment by PatrickF1** [comment]
Can you rebase this on master so we can test more confidently? Thanks!

**Comment by PatrickF1** [comment]
Also, I want to thank you from the bottom of my heart for working on hard on making fzf.fish a better plugin. I've learned a lot from your valuable suggestions and more advanced-shell techniques. I feel bad that our review/feedback/improvement cycles are quite long and also want to give you more of a voice as you've demonstrated your shell prowess, so what do you think about setting up some place where we can chat in real-time? Do you use Gitter or Twitter (or even FB) for chats?

File: functions/__fzf_search_shell_variables.fish (hunk 1 above)
Comment (kidonng) [reply functions/__fzf_search_shell_variables.fish:14]: I was thinking about saving to call `commanline --current-token` twice, but now I think about it again, it's pretty cheap.

**Comment by kidonng** [comment]
> Also, I want to thank you from the bottom of my heart for working on hard on making fzf.fish a better plugin.

I also want to thank you for creating such a amazing plugin and making valuable reviews!

> I feel bad that our review/feedback/improvement cycles are quite long

That's just how open source works 😄 In fact we are quite fast according to my experience.

> so what do you think about setting up some place where we can chat in real-time? Do you use Gitter or Twitter (or even FB) for chats?

I only use Telegram for IM. And let's keep PR related discussion on GitHub.

File: functions/__fzf_search_git_log.fish [hunk 2]
```diff
@@ -10,12 +10,12 @@ function __fzf_search_git_log --description "Search the git log of the current g
...
             # %h gives you the abbreviated commit hash, which is useful for saving screen space, but we will have to expand it later below
             git log --color=always --format=format:'%C(bold blue)%h%C(reset) - %C(cyan)%as%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)' | \
-            fzf --ansi --tiebreak=index --preview='git show --color=always (string split --max 1 " " {})[1]'
+            fzf --ansi --tiebreak=index --preview='git show --color=always (string split --max 1 " " {})[1]' --query=(commandline --current-token)
         )
         if test $status -eq 0
             set abbreviated_commit_hash (string split --max 1 " " $selected_log_line)[1]
             set commit_hash (git rev-parse $abbreviated_commit_hash)
-            commandline --insert $commit_hash
+            commandline --current-token --replace $commit_hash
```
PatrickF1, line 18: Is there really any value to this? I guess it doesn't matter either way but I can't see anyone memorizing git-shas and starting to type them and then triggering this function. Do you do that?

kidonng: - This is consistent with other functions
- You don't have to always type SHA. Let's say I want to `git checkout dev|` (| is the cursor), and I realized I actually want to go to some other commit that mentioned `dev`, so I can trigger it and don't have to type the same thing again.

PatrickF1:
Well, it sure is wonderful to partner with you. What gave you such an interest in this plugin to pour in so much work into it?

Previously, I chatted with @IlanCosman who also helped a ton on Gitter but sure, let's just keep conversations on GitHub since the feedback loop is fast enough already for you.

functions/__fzf_search_git_log.fish
PatrickF1: good point, hadn't considered that use case

functions/__fzf_search_current_dir.fish
```diff
@@ -17,7 +17,7 @@ function __fzf_search_current_dir --description "Search the current directory us
 
         for path in $file_paths_selected
             set escaped_path (string escape "$path")
-            commandline --insert "$escaped_path "
+            commandline --current-token --replace "$escaped_path "
```
PatrickF1, line 20: Sorry to be nitpicky--I know this logic works--but it's a bit hacky. The only reason `commandline --current-token --replace` doesn't keep overwriting itself and instead inserts is because we padded `escaped_path` with a space. I'm currently brainstorming how we can make this code be less deceptive / easier to follow...
One idea off the top of my head is to map `file_paths_selected` using string escape and then use string join on it. Then we can do `commandline --current-token --replace $joined_escaped_paths`

functions/__fzf_search_shell_variables.fish
```diff
@@ -10,11 +10,16 @@ function __fzf_search_shell_variables --description "Search and inspect shell va
     # __echo_value_or_print_message will print an informative message in lieu of the value.
     set variable_name (
         set --names |
-        fzf --preview '__fzf_display_value_or_error {}'
+        fzf --preview '__fzf_display_value_or_error {}' \
+            --query=(commandline --current-token | string replace '$' '')
```
PatrickF1, line 14: We need a comment to explain why you're erasing the $.

functions/__fzf_search_shell_variables.fish
```diff
@@ -10,11 +10,16 @@ function __fzf_search_shell_variables --description "Search and inspect shell va
...
     )
 
     if test $status -eq 0
-        commandline --insert $variable_name
+        if string match -q '$*' (commandline --current-token)
```
PatrickF1, line 18: Need a comment here to explain why there's two cases. Not intuitive.

kidonng:
> Well, it sure is wonderful to partner with you. What gave you such an interest in this plugin to pour in so much work into it?

Because it's easier than writing yet another one... Just kidding. I use it so why not make it better.

functions/__fzf_search_current_dir.fish
kidonng: Wow, I didn't notice that.

functions/__fzf_search_current_dir.fish
```diff
@@ -15,10 +15,7 @@ function __fzf_search_current_dir --description "Search the current directory us
             set file_paths_selected ./$file_paths_selected
         end
 
-        for path in $file_paths_selected
-            set escaped_path (string escape "$path")
-            commandline --insert "$escaped_path "
-        end
+        commandline --current-token --replace (string escape $file_paths_selected | string join ' ')
```
PatrickF1, line 18: `string escape $file_paths_selected` doesn't make them all one giant string? Amazing! I'll double check

functions/__fzf_search_git_log.fish
PatrickF1, line 18: I wonder if we should just allow multi-select here. Don't worry about changing it for this PR, but I'd like to hear your thoughts. It seems you use my plugin even more than me and have a better grasp on what it's good for.

functions/__fzf_search_current_dir.fish
PatrickF1: Amazing, it sure does. That wasn't obvious in the documentation.

PatrickF1 (approved):
Tested and everything looks good. Nice work!

PatrickF1:
Note for posterity: I had struggled with whether to merge this or not because I wasn't sure this change was worth maintaining if it's not very discoverable and it makes fzf.fish's behavior more complex, a bit more unpredictable than always inserting the user's selection. However, I decided to merge because
1) most users will probably have an empty current token when they trigger fzf, so this won't affect them
2) those who do have a current token when they trigger fzf ended up having to go back and modify their commandline because the current token would become one word with what fzf.fish inserts into the commandline
3) there's already a strong precedence of replacing the current token with autocomplete and suggestions being built into fish
4) this behavior can be easily documented with new gifs

functions/__fzf_search_git_log.fish
kidonng: Sorry for the (really) late reply! I intended to get to this earlier but just forgot.
It's fine to keep it as-is, but I don't see any problem to always allow multiple selections. In fact I have `--multi` in my `FZF_DEFAULT_OPTS`.

PatrickF1: Lol how did you find this question after 2.5 months?

kidonng: It's been lying in my inbox along with many others 😅

PatrickF1: I see haha. Well, thanks for getting back to me. I think I'll leave it be without --multi for now.

PR #74, opened 2021-01-04
Use %m-%d for history date format

kidonng:
This is slightly opinionated, but there are two reasons for this change:

- More like ISO 8601 (`YYYY-MM-DD`)
- Reduce ambiguity with paths (no `/`)


PatrickF1 (approved):
Thank you! Yes this is much better. I should have used %d to begin with

Issue #70, opened 2020-12-19
v5 ideas


PatrickF1:
This'll be my scratch space for what I plan to include in the next major release. For anything seeing this, feel free to chime in!

# Configurability
fzf.fish, now that it has reached a broader audience, should cater to more users by being more configurable.

One thing many people have asked for configurability on is the preview file functionality. Areas to make configurable
- ls command for previewing dirs
- bat command for previewing files
- fd command for finding files

# Other
- refactor fzf invocations to use a wrapper that makes sure SHELL and fzf default opts are set, which would also make out-of-the-box Tmux support possible
- fzf to quickly find pid, which can be used with kill, pkill, or cat /proc/:pid/*

paddor:
What about searching for files under a path that has already been typed in the command?

`vi lib/<ctrl-f>`

Or am I just not used to fzf enough yet?

PatrickF1:
Hi @paddor, interesting use-case. That's really clever, but that's also really hard to implement and it'll also require a completely new command since ctrl+f searches the current directory. Let me keep thinking about it.

paddor:
I saw FZF mention this feature as CTRL-T. Maybe you can build on that.

wawa19933:
Just found this exciting plugin and found that it lacks `Alt + C` command to `cd` into selected dir.

PatrickF1:
Hey @wawa19933 sorry to disappoint but I don't plan on implementing that for now. I want to keep the plugin laser focused on completing commands quicker, not executing commands for you. That keeps the code, documentation, learning curve, and maintenance of the plugin (including what features to add or exclude) much easier. However, who knows? Maybe I or someone else will start another fzf plugin that executes commands for you.

PatrickF1:
@wawa19933 actually, there is now something nearly as good as an `Alt+C` command to `cd` into the selected dir. See this commit https://github.com/PatrickF1/fzf.fish/commit/c75fede98374c093466d1a62379bc316c1f3185f

It only requires you to hit enter one extra time after using `Ctrl+F`.

imr0:
Hey, I am bit new to `fish` so the following suggestion might be a bit rough around the edges, but I was looking into the same behaviour that @paddor mentioned because that is the most convenient use case for me as well.
Modifying the first block of the current `__fzf_search_current_dir.fish` as

```
# Token under cursor                                                                       
set --local token (string unescape (commandline --current-token))           
                                                                            
if test -d $token                                                           
[...]

PR #75, opened 2021-01-06
Limit file search to directory under cursor if it has a trailing slash

kidonng:
Derived from https://github.com/PatrickF1/fzf.fish/issues/70#issuecomment-755054602


imr0:
Thanks for the feedback in #70! My preferences for why using completion this way have basically been summarized by @paddor in [#70 (comment)](https://github.com/PatrickF1/fzf.fish/issues/70#issuecomment-755165858), but I will also reiterate them here:

- A lot faster!
- Works with queries containing parent–sibling relationships between directories (i.e. `../` or `./` in the queries).
- More display space for files under the given directory, which looks much cleaner, especially if the preview window is open.
- Can exploit muscle memory for <kbd>TAB</kbd> completion and then invoking `fzf`.
- For—in my opinion—edge cases when the user is searching for a query elsewhere with the exact name of a directory that already exists in the current working directory, hitting <kbd>Ctrl</kbd>+<kbd>f</kbd> first and then typing in the query is still an option.

PatrickF1:
Sorry for the slow turnaround time, been a heck of a day at work...and will continue to be for a while so please be patient.

Anyway, for now, I'm not sold on this. Why? It's not that I don't like the idea. But there are serious downsides when considering the plugin holistically
- The change adds a lot of complexity to the find files function, which I have to maintain. Which I don't want to invest a ton in because I am realizing more and more this problem space of navigating files is best solved by a CLI file manager (e.g. nnn, ranger, midnight commander, lf, Nerdtree, and many many more). It's a very saturated market. I myself am going to move to nnn and probably stop using this feature myself. And I think other people should too. There's no way I can or want to build/maintain anything close to the feature set that nnn provides.
- This is more like adding a quirk to the search files functionality. It's not easily discoverable unless someone looks at the code or follows the releases notes or scans through of the readme. And does anyone really do those things for such a small plugin? I doubt it. I don't want to surprise users with the many quirks of search files, because now there are really 3 modes: search for files with no current token, search for files that replaces the current token, search for files using the current token as a directory.
- Users will probably a hard time remembering and mastering all the quirks of this function. And my goal has always been to KISS so that using the plugin is a no-brainer.
- This needs to be documented and clearly and quickly taught to all users.

Sorry for i'm being harsh. I hate that you already put in the work. But I won't merge it as is. However, I am open to a compromise. Maybe if we use a key bind to reload fzf's input list with the contents of the directory in the query? e.g. `fzf --bind ctrl-f:reload(fd $current_token)`?

kidonng:
> * For—in my opinion—edge cases when the user is searching for a query elsewhere with the exact name of a directory that already exists in the current working directory, hitting Ctrl+f first and then typing in the query is still an option.

We can add a little check here. If there is no other match in the currect directory, we will use the entered directory as a base.

kidonng:
> * The change adds a lot of complexity to the find files function, which I have to maintain.

I don't think this is *that* complex, but yeah, I'm not the one to say it.

> navigating directories is best solved by a CLI file manager

We are not building a file manager. This is a file picker after all, any only recently it got addition so it can be used for quickly navigating to folders.

Third-party program can't intergrate with Fish at this level (unless you would like to keep the program open), which means this function will never become useless.

> 3 modes: search for files with no current token,

That is not a mode, we only save the user from typing `./` themselves.

I agree with the rest of narration.

kidonng:
I just made a change which gives a compelling reason why this will be a nice addition with no cost.

The base directory will only be changed to the token under cursor if:

- The token under cursor is a directory
- **The token ends with slash (`/`)**

Why I call this a compelling reason? Because when you search with `src/`, you are definitely searching inside `src`, not currect directory.

functions/__fzf_search_current_dir.fish
```diff
@@ -2,10 +2,22 @@ function __fzf_search_current_dir --description "Search the current directory us
...
     # See similar comment in __fzf_search_shell_variables.fish.
     set --local --export SHELL (command --search fish)
-    set file_paths_selected (
-        fd --hidden --color=always --exclude=.git 2>/dev/null |
-        fzf --multi --ansi --preview='__fzf_preview_file {}' --query=(commandline --current-token)
-    )
+
+    set fd_arguments --hidden --color=always --exclude=.git
+    set fzf_arguments --multi --ansi --preview='__fzf_preview_file {}'
+    set token (commandline --current-token | string unescape)
```
PatrickF1, line 8: in what cases would you have to string unescape? if the current token is "folder name"?

PatrickF1:
Just tested. The file previews broke :(
Otherwise I like the idea and will definitely merge once it's passed testing.

functions/__fzf_search_current_dir.fish
imr0: This can happen if the folder name has escaped whitespace or other special characters (e.g. braces) that come from fish TAB completion.

kidonng:
@PatrickF1 Preview fixed.

kidonng:
Yeah I kinda messed up my local Git history when I merged master and did my first force push.

PatrickF1 (approved):
Nice work!

PatrickF1 (changes requested):
Sorry it still doesn't work. I would have thought the fix would work but now I don't know what's wrong.
Screen Shot 2021-01-08 at 2 21 40 PM

kidonng:
Works on my machine™️

image

PatrickF1 (approved):
Ah you're right--it works again now--not sure what I did wrong.

kidonng:
Just noticed fzf does this in their own Fish plugin: https://github.com/junegunn/fzf#fish-shell

PatrickF1:
Nice, I'm glad we added it though it's causing quite a bit of maintenance and documentation pain.

PR #76, opened 2021-01-07
Allow quick open with pre-populated query

kidonng:
This PR improves both #72 and #73.

- Also prepends `./` to the only selected path if it is a file. This has two benefits:
	- Prepending `./` allows you to open a executable, just like entering a directory. (But the latter is Fish magic, while the former is not)
	- This makes the feature more consistent. The user doesn't have to remember that they had to select a folder to make it work, now they only needs to know "If I only made one selection, I can quickly open it!".
- #73 doesn't play nice with #72, because we assume the commandline is empty. We are now using a smarter check: if the length of the whole commandline matches the length of the first token of the commandline, we can assume the user is making use of the quick open feature. So #72 will work with pre-populated query.


PatrickF1:
> This makes the feature more consistent. The user doesn't have to remember that they had to select a folder to make it work, now they only needs to know "If I only made one selection, I can quickly open it!".

Not quite. Well, as you said it only works with executables. What about for text files which is probably the most common use case? I'm indifferent though, will probably still merge this after some more thought.

kidonng:
> Well, as you said it only works with executables. What about for text files which is probably the most common use case?

That's why we are lifting the limit while still being generic: we don't prepend `$EDITOR` or something like that.

My example of executables may not be good, but I hope you can get what I mean. It's literally just, if you only selected one path, we prepend `./` for you, which you can make use of.

functions/__fzf_search_current_dir.fish
```diff
@@ -8,10 +8,18 @@ function __fzf_search_current_dir --description "Search the current directory us
...
+        # If the path is an executable, the user can hit Enter one more time to execute it.
+        # If the path is a directory, the user can hit Enter one more time to quickly cd into it, because fish will
+        # attempt to cd implicitly if a directory name starts with a dot.
+        set first_token_length (string length (commandline --tokenize)[1])
+
+        if test -z "$first_token_length"
+            set first_token_length 0
+        end
+
+        if test (string length (commandline)) = $first_token_length && test (count $file_paths_selected) = 1
```
PatrickF1, line 22: Can you reverse the order of the if statement for performance reasons?

PatrickF1: Also, can't you just `test (count (commandline --tokenize) = 1)` to tell if the user only inputted one token?

kidonng: > Can you reverse the order of the if statement for performance reasons?

As you wish, but two builtins vs one builtin doesn't make any difference.

kidonng: > just `test (count (commandline --tokenize) = 1)` to tell if the user only inputted one token

- This doesn't allow it being triggered with an empty command line
- `<command> |` (`|` indicates cursor position) is considered one token by `commandline --tokenize`, which is not desirable.

PatrickF1: Oh good point, forgot commandline was a builtin too.

I still find the logic convoluted though. If all we want to know is if the user is still typing the first token, can't we just compare (commandline) to (commandline --current-token)?

PatrickF1: Or even `test (commandline --tokenize) = (commandline --current-token)`

PatrickF1:
@kidonng want me to finish this for you?

functions/__fzf_search_current_dir.fish
kidonng: > `test (commandline --tokenize) = (commandline --current-token)`

Brilliant! We can even drop `--tokenize`.

PatrickF1: Actually, we should keep tokenize for situations where the user accidentally included a whitespace in the front or back for whatever reason but still has the cursor on the current token. I'll add it in and test.

PatrickF1: Oh darn, now it doesn't work if the input is empty. Hmm...

PatrickF1: Oh it's because if the commandline is empty, then (commandline --tokenize) is empty and test doesn't work
```
test: Missing argument at index 2

~/.config/fish/functions/__fzf_search_current_dir.fish (line 29):
        if test (count $file_paths_selected) = 1 && test ""(commandline --tokenize) = ""(commandline --current-token)
                                                    ^
in function '__fzf_search_current_dir'
```

functions/__fzf_search_current_dir.fish
```diff
@@ -21,11 +21,17 @@ function __fzf_search_current_dir --description "Search the current directory. R
...
-        if test (count (commandline --tokenize)) = 0 && test (count $file_paths_selected) = 1 && test -d $file_paths_selected
-            set file_paths_selected ./$file_paths_selected
+        # If the user was in the middle of inputting the first token and only one path is selected,
+        # then prepend ./ to the selected path so that
+        # - if the path is an executable, the user can hit Enter one more time to execute it.
+        # - if the path is a directory, the user can hit Enter one more time to quickly cd into it (fish will
+        #   attempt to cd implicitly if a directory name starts with a dot)
+        if test (count $file_paths_selected) = 1
+            set commandline_tokens (commandline --tokenize)
+            set current_token (commandline --current-token)
```
PatrickF1, line 31: @kidonng please LMK if you have a better solution to command substitution resulting in an empty string. Seems fish doesn't allow you to do "(commandline --tokenize)" so I have to use variables.

PatrickF1 (approved):

functions/__fzf_search_current_dir.fish
kidonng: Yes, you have to use a variable (which IMO is not bad), see https://fishshell.com/docs/current/index.html#cartesian-products.

kidonng:
Thanks for your work on this! ✨

PatrickF1:
Thanks for getting it started @kidonng!

PR #80, opened 2021-01-14
Prevent arguments from being interpreted as options

kidonng:
User inputs, file names and commands can contain leading hypens (`--`), which may cause errors or the function not working.

Examples where fzf.fish would error:
- the cursor is over the token --something when the user executes find file
- the selected changed path from git status is -folder
- the selected command is --function

The solution is to delineate the end of options with --, which is a POSIX specification defined in https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02 (guideline 10), which is respected by Fish builtins. After --, the command will always interpret the arguments as positional arguments.

### DIFF
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -9,7 +9,7 @@ function __fzf_search_current_dir --description "Search the current directory. R
 
     # If the current token is a directory and has a trailing slash,
     # then use it as fd's base directory.
-    if string match --quiet "*/" $token && test -d $token
+    if string match --quiet -- "*/" $token && test -d $token
         set --append fd_arguments --base-directory=$token
         # use the directory name as fzf's prompt to indicate the search is limited to that directory
         set --append fzf_arguments --prompt=$token --preview="__fzf_preview_file $token{}"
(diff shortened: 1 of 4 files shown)


PatrickF1 (approved):
Wow some SQL-injection-like bugs you're fixing here. Amazing, didn't even consider these edge cases!

PatrickF1:
Can you link me to documentation that describes how -- fixes it?

PatrickF1:
Doesn't __fzf_search_shell_variables.fish have the same problem?

kidonng:
It's included.

PatrickF1:
I really don't see it...

PatrickF1:
@kidonng Oh LOL I typoed. You probably thought I was crazy. I meant __fzf_search_git_status.fish

kidonng:
Right, I forgot that it will insert paths eventually.

PatrickF1:
Nice, thanks! I was about to ask why you didn't add `--` to all the string commands but then I realized all the lines will always start with a space and a letter (e.g. ` M functions/file.fish` so there's no need. Good thing you wrote the code first!

PR #81, opened 2021-01-15
Fix bug in extracting variable values containing |

PatrickF1:
The bug happens whenever the variable being previewed contains `|`. What happens is that the third regex used in `__fzf_extract_var_info` is greedy and will over-remove text up until the second to last `|`. For example:
```
$ set variable "a | b | c | d"
$ __fzf_extract_var_info variable (set --show | psub)
set in global scope, unexported, with 1 elements
[1]  d
```

With these changes, it now works properly:
```
$ set variable "a | b | c | d"
$ __fzf_extract_var_info variable (set --show | psub)
set in global scope, unexported, with 1 elements
[1] a | b | c | d
```

Additionally, I've added a Fishtape test to make debugging these tricky regexes more easily in the future, which I expect to need to do as fish shell on master has changed the output of `set --show`.

### DIFF
--- a/functions/__fzf_extract_var_info.fish
+++ b/functions/__fzf_extract_var_info.fish
@@ -10,7 +10,7 @@ function __fzf_extract_var_info --argument-names variable_name set_show_output -
     # $variable_name: set in global scope
     # ...with...
     # set in global scope
-    string replace --regex "^\\\$$variable_name: " "" |
+    string replace --regex "^\\\$$variable_name: " '' |
 
     # From the lines of values, keep only the index and value, replacing...
     # $variable_name[1]: length=14 value=|variable_value|
--- a/functions/__fzf_extract_var_info.fish
+++ b/functions/__fzf_extract_var_info.fish
@@ -6,17 +6,17 @@ function __fzf_extract_var_info --argument-names variable_name set_show_output -
     # $variable_name[
     string match --entire --regex "^\\\$$variable_name(?:: set|\[)" <$set_show_output |
 
-    # Strip the variable name from the variable info, replacing...
+    # Strip the variable name from the scope info, replacing...
     # $variable_name: set in global scope
     # ...with...
     # set in global scope
-    string replace --regex "^\\\$$variable_name: " '' |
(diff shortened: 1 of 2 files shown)


functions/__fzf_extract_var_info.fish
```diff
@@ -6,17 +6,17 @@ function __fzf_extract_var_info --argument-names variable_name set_show_output -
     # $variable_name[
     string match --entire --regex "^\\\$$variable_name(?:: set|\[)" <$set_show_output |
```
PatrickF1, line 7: Self-note for later: why are 3 `\`s necessary to escape the first `$`? Well, the last one is needed to prevent fish from starting a variable substitution. Then, to prevent the regex library from interpreting the `$` as an anchor, we need to prefix it with a `\`. But to prevent that `\` from applying to the now-literal `$`, we need to escape it as well with another `\`.

functions/__fzf_extract_var_info.fish
```diff
@@ -6,17 +6,17 @@ function __fzf_extract_var_info --argument-names variable_name set_show_output -
...
     # set in global scope
-    string replace --regex "^\\\$$variable_name: " '' |
+    string replace --regex "^\\\$$variable_name: " "" |
 
     # From the lines of values, keep only the index and value, replacing...
     # $variable_name[1]: length=14 value=|variable_value|
     # ...with...
     # [1] variable_value
-    string replace --regex "^\\\$$variable_name(\[.+\]).+\|(.+)\|\$" '\$1 \$2'
+    string replace --regex "^\\\$$variable_name(\[\d+\]).+?\|(.+)\|\$" '\$1 \$2'
```
PatrickF1, line 19: The bug fix is here: make the `.+` that comes after the first capturing group non-greedy by appending a `?`.

A minor unrelated change is using `\d` to match inside the [ ] to make it more obvious that it's supposed to be index numbers.

kidonng:
Just noticed this in the release notes. I remember using non-greedy matching sometime in the original PR, but it seems to be lost. Anyway, thanks for fixing it!

PR #109, opened 2021-01-30
Expand tilde while searching files

### LINKED ISSUE #88: file paths doesn't work with tilde ~ prefix
For example `vim ~/<ctrl-f>` does not work. When I expand the path (`vim /Users/aaa/<ctrl-f>`) it works fine. It would be great if this feature could be added.

Using fish 3.1.2 and fzf.fish 5.3

kidonng:
Fix #88 

Fish's `test` and `fd` doesn't expand `~` automatically, so we need to expand it ourselves.

image

### DIFF
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -6,11 +6,12 @@ function __fzf_search_current_dir --description "Search the current directory. R
     set fd_arguments --hidden --color=always --exclude=.git
     set fzf_arguments --multi --ansi
     set token (commandline --current-token | string unescape)
+    set expanded_token (string replace --regex -- "^~/" $HOME/ $token)
 
     # If the current token is a directory and has a trailing slash,
     # then use it as fd's base directory.
-    if string match --quiet -- "*/" $token && test -d $token
-        set --append fd_arguments --base-directory=$token
+    if string match --quiet -- "*/" $token && test -d $expanded_token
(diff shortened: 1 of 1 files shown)


functions/__fzf_search_current_dir.fish
```diff
@@ -6,11 +6,12 @@ function __fzf_search_current_dir --description "Search the current directory. R
     set fd_arguments --hidden --color=always --exclude=.git
     set fzf_arguments --multi --ansi
     set token (commandline --current-token | string unescape)
+    set expanded_token (string replace --regex -- "^~/" $HOME/ $token)
```
kidonng, line 9: In fact we can use a unquoted `~` to replace `$HOME`, but that would be too confusing 😄

jorgebucaran, line 9: I personally find `string replace --regex -- "^~/" ~/ $token` more aesthetically pleasing, but this works too. 👍

jorgebucaran (approved):

functions/__fzf_search_current_dir.fish
jorgebucaran: I personally find `string replace --regex -- "^~/" ~/ $token` more aesthetically pleasing, but this works too.

PatrickF1:
Nice work! Thanks @kidonng and thanks for helping review @jorgebucaran :)

PR #110, opened 2021-01-30
Output to stdout when executing a command substitution

kidonng:
See https://github.com/PatrickF1/fzf.fish/discussions/104#discussioncomment-322968

I left out `__fzf_search_history` because it doesn't make much sense.

### DIFF
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -34,7 +34,12 @@ function __fzf_search_current_dir --description "Search the current directory. R
             end
         end
 
-        commandline --current-token --replace -- (string escape -- $file_paths_selected | string join ' ')
+        set output (string escape -- $file_paths_selected | string join ' ')
+        if status is-command-substitution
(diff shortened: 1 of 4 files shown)


IlanCosman:
Could you explain the use case for running this plugin non-interactively? I'm don't quite understand the motivation.

PatrickF1:
Hey @IlanCosman let's move discussion to the...discussions please! I'll comment back there

IlanCosman:
Can we extract

```fish
if status is-command-substitution
    printf $output
else
    commandline --current-token --replace -- $output
end
```

to a function? It comes up super often.

kidonng:
Of course, but I don't have any idea for naming the function.

PatrickF1:
Let's put a hold on this for now...still undecided about if I want to do this or not. Thanks

kidonng:
Yeah, this is just proof of concept.

kidonng:
Close it for the moment since #140 meets most customization needs

PR #113, opened 2021-01-31
Refactor: one folder per test suite, one file per test case

PatrickF1:
1. Put each test in its own file. From fishtape docs
> Each test file runs inside its own shell, so you can modify the global environment without cluttering your session or breaking other tests

2. Stopped making all the variables local, since I realized they won't pollute the session running the tests.
3. Replaced fish version badge with build status badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # fzf.fish 🔍🐟
 
-[![latest release badge][]][releases] [![fish version badge][]](#installation) [![awesome badge][]][awesome fish]
+[![latest release badge][]][releases] [![build status badge][]][actions] [![awesome badge][]][awesome fish]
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 # fzf.fish 🔍🐟
 
-[![latest release badge][]][releases] [![fish version badge][]](#installation) [![awesome badge][]][awesome fish]
+[![latest release badge][]][releases] [![build status badge][]][actions] [![awesome badge][]][awesome fish]
```
PatrickF1, line 5: I figured seeing the build status is more helpful and assuring than the fish version. People likely have the latest fish anyway and the installation instructions make it clear what version is necessary.

PR #114, opened 2021-02-01
Automate smoke tests for every feature

PatrickF1:
While I can't and don't plan on trying to test every possible code path, I can at least set up smoke tests. These smoke tests should catch many cases where a regression would completely break a feature, and thus prevent regressions from getting merged in. So while it won't obviate manual testing, it will cut down on manual testing for both me and anyone who wants to contribute to fzf.fish!

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -23,6 +23,6 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     else if test -p "$file_path"
         __fzf_report_file_type "$file_path" "named pipe"
     else
-        echo "Unknown file type." >&2
+        echo "File doesn't exist." >&2
     end
 end
--- a/tests/search_current_dir/expands_~_base_dir.fish
+++ b/tests/search_current_dir/expands_~_base_dir.fish
@@ -1,8 +1,11 @@
-# mock everything so that the only thing that gets sent to stdout are the args passed into fd
-mock fd \* "echo \$argv"
+set --export fd_args
+function fd
+    set fd_args $argv
+    return 0
+end
 mock fzf \* ""
 mock commandline --current-token "echo ~/"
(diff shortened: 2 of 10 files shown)


tests/search_current_dir/expands_~_base_dir.fish
```diff
@@ -0,0 +1,8 @@
+# mock everything so that the only thing that gets sent to stdout are the args passed into fd
+mock fd \* "echo \$argv"
+mock fzf \* ""
+mock commandline --current-token "echo ~/"
+mock commandline \* ""
+set fd_args (__fzf_search_current_dir)
+@test "~/ not in fd args" -z (string match "~/" $fd_args)
+@test "~/ is expanded to $HOME" -n (string match --entire "$HOME" $fd_args)
```
jorgebucaran, line 8: As a coding convention, I always specify `--`:
```fish
string match --entire -- "$HOME" $fd_args
```

jorgebucaran (approved):

tests/search_current_dir/expands_~_base_dir.fish
PatrickF1: Thanks Jorge! That's an important convention I will copy.

PatrickF1:
Thanks for approving. I'm actually gonna add a lot more tests. Would appreciate another review later 😉

functions/__fzf_preview_file.fish
```diff
@@ -23,6 +23,6 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     else if test -p "$file_path"
         __fzf_report_file_type "$file_path" "named pipe"
     else
-        echo "Unknown file type." >&2
+        echo "File doesn't exist." >&2
```
PatrickF1, line 26: Fixed a small bug: realized the only time this happens is if the file doesn't exist.

PR #118, opened 2021-02-05
Grep current directory with ripgrep

ovv:
Hello :wave: , great plugin !

Before I polish this would you consider merging something like this ?

### DIFF
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -8,6 +8,7 @@ set __fzf_search_vars_cmd '__fzf_search_shell_variables (set --show | psub) (set
 if not set --query fzf_fish_custom_keybindings
     # \cf is Ctrl+f
     bind \cf '__fzf_search_current_dir'
+    bind \cg '__fzf_grep_current_dir'
     bind \cr '__fzf_search_history'
     bind \cv $__fzf_search_vars_cmd
     # The following two key binding use Alt as an additional modifier key to avoid conflicts
@@ -17,6 +18,7 @@ if not set --query fzf_fish_custom_keybindings
     # set up the same key bindings for insert mode if using fish_vi_key_bindings
     if test "$fish_key_bindings" = 'fish_vi_key_bindings'
         bind --mode insert \cf '__fzf_search_current_dir'
+        bind --mode insert \cg '__fzf_grep_current_dir'
         bind --mode insert \cr '__fzf_search_history'
         bind --mode insert \cv $__fzf_search_vars_cmd
         bind --mode insert \e\cl '__fzf_search_git_log'
--- a/functions/__fzf_grep_current_dir.fish
+++ b/functions/__fzf_grep_current_dir.fish
@@ -1,0 +1,40 @@
+function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish to execute __fzf_preview_file.
+    # See similar comment in __fzf_search_shell_variables.fish.
+    set --local --export SHELL (command --search fish)
+
+    set rg_cmd rg --ignore-case --files-with-matches
+    set redirect "< /dev/tty > /dev/tty 2>&1"
+
+    set token (commandline --current-token | string unescape)
+
+    set fzf_arguments --multi --ansi --exact --phony --query=$token \
+        --preview='rg --ignore-case --pretty --context 10 {q} {}' \
+        # reload the search when the query cange
+        --bind "change:reload:$rg_cmd {q} || true" \
+        # launch vim on ctrl-e, if there is a query start a search in vim
+        --bind "ctrl-e:execute(if test {q}; vim -c "/{q}" {} $redirect; else; vim {} $redirect; end)" \
+        # launch less on ctrl-v, if there is a query start a search in less
+        --bind "ctrl-v:execute(if test {q}; less -i -p "{q}" {} $redirect; else; less {} $redirect; end)" \
+
+    set file_paths_selected ($rg_cmd $token | fzf $fzf_arguments)
+
+    if test $status -eq 0
+        # If the user was in the middle of inputting the first token and only one path is selected,
+        # then prepend ./ to the selected path so that
+        # - if the path is an executable, the user can hit Enter one more time to immediately execute it
+        # - if the path is a directory, the user can hit Enter one more time to immediately cd into it (fish will
+        #   attempt to cd implicitly if a directory name starts with a dot)
+        if test (count $file_paths_selected) = 1
+            set commandline_tokens (commandline --tokenize)
+            set current_token (commandline --current-token)
+            if test "$commandline_tokens" = "$current_token"
+                set file_paths_selected ./$file_paths_selected
+            end
+        end
+
+        commandline --current-token --replace -- (string escape -- $file_paths_selected | string join ' ')
+    end
+
+    commandline --function repaint
+end
--- a/functions/__fzf_grep_current_dir.fish
+++ b/functions/__fzf_grep_current_dir.fish
@@ -2,18 +2,17 @@ function __fzf_grep_current_dir --description "Search current directory content 
     # Make sure that fzf uses fish as shell.
     set --local --export SHELL (command --search fish)
 
-    set rg_cmd rg --color=always --line-number --smart-case --no-heading
+    set rg_cmd rg --color=always --smart-case --no-heading --count --auto-hybrid-regex
     set token (commandline --current-token | string unescape)
 
-    set fzf_arguments --multi --ansi --delimiter=: --with-nth=1,3 --disabled --query=$token \
-        --preview 'bat --style=numbers --color=always --highlight-line {2} {1}' \
-        # center the preview window on the highlighted line
-        --preview-window +{2}-/2 \
+    set fzf_arguments --multi --ansi --disabled --delimiter=: --query=$token \
+        --preview='rg --smart-case --auto-hybrid-regex --pretty --context 2 {q} {1}' \
         # reload the search when the query change
         --bind "change:reload:$rg_cmd {q} || true"
 
+    set file_paths_selected
     for file in ($rg_cmd $token | fzf $fzf_arguments)
-        set file_paths_selected $file_paths_selected (string split : $file)[1]
+       set --append file_paths_selected (string split : $file)[1]
     end
 
     if test $status -eq 0
--- a/functions/__fzf_grep_current_dir.fish
+++ b/functions/__fzf_grep_current_dir.fish
@@ -1,22 +1,20 @@
 function __fzf_grep_current_dir --description "Search current directory content for current token. Replace the current token with the selected file paths."
-    # Make sure that fzf uses fish.
+    # Make sure that fzf uses fish as shell.
     set --local --export SHELL (command --search fish)
 
-    set rg_cmd rg --ignore-case --files-with-matches
-    set redirect "< /dev/tty > /dev/tty 2>&1"
-
+    set rg_cmd rg --color=always --line-number --smart-case --no-heading
     set token (commandline --current-token | string unescape)
 
-    set fzf_arguments --multi --exact --disabled --query=$token \
-        --preview='rg --smart-case --pretty --context 10 {q} {}' \
-        # reload the search when the query cange
-        --bind "change:reload:$rg_cmd {q} || true" \
-        # launch vim on ctrl-e, if there is a query start a search in vim
-        --bind "ctrl-e:execute(if test {q}; vim -c "/{q}" {} $redirect; else; vim {} $redirect; end)" \
-        # launch less on ctrl-v, if there is a query start a search in less
-        --bind "ctrl-v:execute(if test {q}; less -i -p "{q}" {} $redirect; else; less {} $redirect; end)" \
+    set fzf_arguments --multi --ansi --delimiter=: --with-nth=1,3 --disabled --query=$token \
+        --preview 'bat --style=numbers --color=always --highlight-line {2} {1}' \
+        # center the preview window on the highlighted line
+        --preview-window +{2}-/2 \
+        # reload the search when the query change
+        --bind "change:reload:$rg_cmd {q} || true"
 
-    set file_paths_selected ($rg_cmd $token | fzf $fzf_arguments)
+    for file in ($rg_cmd $token | fzf $fzf_arguments)
+        set file_paths_selected $file_paths_selected (string split : $file)[1]
+    end
 
     if test $status -eq 0
         commandline --current-token --replace -- (string escape -- $file_paths_selected | string join ' ')
--- a/functions/__fzf_grep_current_dir.fish
+++ b/functions/__fzf_grep_current_dir.fish
@@ -1,6 +1,5 @@
-function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
-    # Make sure that fzf uses fish to execute __fzf_preview_file.
-    # See similar comment in __fzf_search_shell_variables.fish.
+function __fzf_grep_current_dir --description "Search current directory content for current token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish.
     set --local --export SHELL (command --search fish)
 
     set rg_cmd rg --ignore-case --files-with-matches
@@ -8,8 +7,8 @@ function __fzf_grep_current_dir --description "Search current directory content 
 
     set token (commandline --current-token | string unescape)
 
-    set fzf_arguments --multi --ansi --exact --phony --query=$token \
-        --preview='rg --ignore-case --pretty --context 10 {q} {}' \
+    set fzf_arguments --multi --exact --disabled --query=$token \
+        --preview='rg --smart-case --pretty --context 10 {q} {}' \
         # reload the search when the query cange
         --bind "change:reload:$rg_cmd {q} || true" \
         # launch vim on ctrl-e, if there is a query start a search in vim
@@ -20,19 +19,6 @@ function __fzf_grep_current_dir --description "Search current directory content 
     set file_paths_selected ($rg_cmd $token | fzf $fzf_arguments)
 
     if test $status -eq 0
-        # If the user was in the middle of inputting the first token and only one path is selected,
-        # then prepend ./ to the selected path so that
-        # - if the path is an executable, the user can hit Enter one more time to immediately execute it
-        # - if the path is a directory, the user can hit Enter one more time to immediately cd into it (fish will
-        #   attempt to cd implicitly if a directory name starts with a dot)
-        if test (count $file_paths_selected) = 1
-            set commandline_tokens (commandline --tokenize)
-            set current_token (commandline --current-token)
-            if test "$commandline_tokens" = "$current_token"
-                set file_paths_selected ./$file_paths_selected
-            end
-        end
-
         commandline --current-token --replace -- (string escape -- $file_paths_selected | string join ' ')
     end
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,12 @@ Use `fzf.fish` to interactively find and insert into the command line:
   - ignores files that are also ignored by git
   - <kbd>Tab</kbd> to multi-select
 
+### Files containing a pattern
+
+- **Search input:** all files containing the current query
+- **Key binding and mnemonic:** <kbd>Ctrl</kbd>+<kbd>G</kbd> (`G` for grep)
+- **Preview window:** the file with occurrences of the query highlighted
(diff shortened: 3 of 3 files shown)


PatrickF1:
Hey @ovv. Thanks! And thank you for the great idea. TBH I've had the same idea before but didn't know how to properly fit it in with the rest of the features. The UX for it is kind of unintuitive--you type in a token to use to grep for and aren't actually going to use in the command. Why not just ask straight prompt the user for what he wants to grep for instead? What do you think?

ovv:
> you type in a token to use to grep for and aren't actually going to use in the command

I'm not sure I fully understand what you mean. There are two cases

1. There is a token on the command line: we grep for it and display the result
2. There is no token: we display all files in the directory

For both cases you can then update your query and the result will update

PatrickF1:
> There is a token on the command line: we grep for it and display the result

Yeah I get that. Think through the workflow. It's not natural. Say I want to edit a file containing the word fish. I wouldn't do ` vim fish<execute function>`. I would rather do `vim <execute function and type in fish>`.

> There is no token: we display all files in the directory

How is case 2 any different than the __fzf_search_current_dir?

PatrickF1:
Oh sorry I you put in a `change:reload:rg --ignore-case --files-with-matches --no-ignore-vcs {q} || true`. I didn't notice that. So basically the idea is that whatever gets put into fzf query box, that is used to do `--file-with-matches` and grep for occurences in each file. Very nice! I like this very much. Let me sit on it some more to think about perfecting the behavior.

PatrickF1:
Darn, fzf has no support for debounce https://github.com/junegunn/fzf/issues/965. My main concerns right now are
1) performance: in a big repo, not having debounce and typing a 10 char string could kill someone's computer, though it is their choice
2) unintuitive to use; how do we easily convey to a user who doesn't read the readme carefully that this feature doesn't filter based on file name but based on their contents?

Any ideas would be appreciated.

ovv:
Full disclosure I took that line from: https://github.com/junegunn/fzf/wiki/examples#searching-file-contents

> performance: in a big repo, not having debounce and typing a 10 char string could kill someone's computer, though it is their choice

This is where typing the token before triggering the function is useful. I agree that it could be better but that's one of the downside of linking tools together instead of something more integrated. We should keep an eye out if that feature is added to fzf at some point. 

Note that I disabled searching in hidden files/folder for performance reason. This could potentially be configurable.

> unintuitive to use; how do we easily convey to a user who doesn't read the readme carefully that this feature doesn't filter based on file name but based on their contents?

Well having both features make it more obvious that they are different. 

I'm a bit bummed that the name `__fzf_search_current_dir` is already taken because I feel it's more appropriate for this feature. Thought we can't really change it without breaking some user setup. On the other hand `grep` in the name is even more explicit.

PatrickF1:
> Note that I disabled searching in hidden files/folder for performance reason.

Very smart and it makes a ton of sense, since most of the time people will be looking for code, which doesn't live in hidden directories.

> I'm a bit bummed that the name __fzf_search_current_dir is already taken because I feel it's more appropriate for this feature.

I'm not closed to renaming __fzf_search_current_dir (I may soon abstract away all the function names are variables). What would you change its name to, though?

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,36 @@
+function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish to execute __fzf_preview_file.
+    # See similar comment in __fzf_search_shell_variables.fish.
+    set --local --export SHELL (command --search fish)
+
+    set token (commandline --current-token | string unescape)
+
+    set fzf_arguments --multi --ansi --exact --phony --reverse --query=$token \
```
PatrickF1, line 8: What's the point of both `--exact --phony`? Also `--reverse` is already in the default args.

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,36 @@
...
+        --preview='rg --ignore-case --pretty --context 10 {q} {}' \
+        --bind "change:reload:rg --ignore-case --files-with-matches --no-ignore-vcs {q} || true"
+
+    if test "$token" = ""
+        set file_paths_selected (rg --files | fzf $fzf_arguments)
+    else
+        set file_paths_selected (rg --files-with-matches $token | fzf $fzf_arguments)
+    end
```
PatrickF1, line 16: This might not be possible, but I'd really like us to not have to make a special case search for starting the search using the token. If only there was a way to automatically execute `reload` as soon as fzf starts up 🤔

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,36 @@
+function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish to execute __fzf_preview_file.
+    # See similar comment in __fzf_search_shell_variables.fish.
+    set --local --export SHELL (command --search fish)
+
+    set token (commandline --current-token | string unescape)
+
+    set fzf_arguments --multi --ansi --exact --phony --reverse --query=$token \
+        --preview='rg --ignore-case --pretty --context 10 {q} {}' \
+        --bind "change:reload:rg --ignore-case --files-with-matches --no-ignore-vcs {q} || true"
```
PatrickF1, line 10: why `--no-ignore-vcs`?

functions/__fzf_grep_current_dir.fish
PatrickF1: Btw `rg --files-with-matches ""` gives the same output as `rg --files`, so you don't have to do two cases based on the token.

ovv:
> I'm not closed to renaming __fzf_search_current_dir (I may soon abstract away all the function names are variables). What would you change its name to, though?

I would rename it to `__fzf_list_current_dir` or `__fzf_ls_current_dir`

If you go down that path I would also like `conf.d/fzf.fish` to be renamed to `conf.d/_fzf.fish`. I like to have my custom configurations names after the package but this would overwrite the provided configuration.

functions/__fzf_grep_current_dir.fish
ovv: I believe `--exact --phony` is needed to let ripgrep do the filtering. Here all filtering is provided by ripgrep, fzf is only a "GUI" on top.

functions/__fzf_grep_current_dir.fish
ovv: Nice find, that would simplify things a bit. I'll push an update tomorrow

functions/__fzf_grep_current_dir.fish
ovv: Yeah this should be removed.

I believe it would be best to allow user to configure some default argument to ripgrep

ovv:
Updated following your recommendations, I also added two `--bind` so you can open `vim`/`less` from fzf to get a better view of the file.

Let me know what else is needed

PatrickF1:
Wow you are an fzf scripting master! Can't wait to get this perfected and merged!

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
+function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
```
kidonng, line 1: Escape `$token`?

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
...
+    # Make sure that fzf uses fish to execute __fzf_preview_file.
+    # See similar comment in __fzf_search_shell_variables.fish.
+    set --local --export SHELL (command --search fish)
+
+    set rg_cmd rg --ignore-case --files-with-matches
+    set redirect "< /dev/tty > /dev/tty 2>&1"
+
+    set token (commandline --current-token | string unescape)
+
+    set fzf_arguments --multi --ansi --exact --phony --query=$token \
```
kidonng, line 11: Better remove `--ansi` because there's no color (and `rg --color=always` gives a single color, which isn't useful too).

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
+function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish to execute __fzf_preview_file.
+    # See similar comment in __fzf_search_shell_variables.fish.
+    set --local --export SHELL (command --search fish)
+
+    set rg_cmd rg --ignore-case --files-with-matches
```
kidonng, line 6: I would suggest using `--smart-case` instead of `--ignore-case`.

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
...
+    set redirect "< /dev/tty > /dev/tty 2>&1"
+
+    set token (commandline --current-token | string unescape)
+
+    set fzf_arguments --multi --ansi --exact --phony --query=$token \
+        --preview='rg --ignore-case --pretty --context 10 {q} {}' \
+        # reload the search when the query cange
+        --bind "change:reload:$rg_cmd {q} || true" \
+        # launch vim on ctrl-e, if there is a query start a search in vim
+        --bind "ctrl-e:execute(if test {q}; vim -c "/{q}" {} $redirect; else; vim {} $redirect; end)" \
```
kidonng, line 16: Use `$EDITOR` instead of `vim`?

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
...
+        # launch less on ctrl-v, if there is a query start a search in less
+        --bind "ctrl-v:execute(if test {q}; less -i -p "{q}" {} $redirect; else; less {} $redirect; end)" \
```
kidonng, line 18: Like above, use `$PAGER`?

kidonng:
Throw in my two cents: I have a funcion that uses ripgrep, fzf and bat for full text fuzzy search:

```fish
# Based on https://github.com/junegunn/fzf/issues/1598#issuecomment-719573480
function rzf
    rg --color=always --line-number --smart-case "$argv" |
    fzf --ansi --delimiter : \
      --with-nth=1,3 \
      --preview 'bat --style=numbers --color=always --highlight-line {2} {1}' \
      --preview-window +{2}-/2 |
    string split : -f 1
end
```

This is not the same as what this PR is achieving, but IMO more convenient to use.

image

PatrickF1:
@kidonng The function you posted is meant to by triggered with an argument, right? So it's not meant to be used in the same way.

I love the idea of using bat though. Since bat is already a dep, we should use that to preview the occurrences of the matching string in the file.

kidonng:
> @kidonng The function you posted is meant to by triggered with an argument, right?

It's optional, as can be seen in the screenshot. It can be used to narrow down the initial results, which is also useful for large folder.

functions/__fzf_grep_current_dir.fish
PatrickF1: As much as I love these changes, let's keep the extra key bindings out for now. Main reason is because __fzf_search_current_dir doesn't have them. I'll either add them later in one fell swoop later or just teach users to add them themselves so they can customize it to their hearts content. Thanks!

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
+function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish to execute __fzf_preview_file.
```
PatrickF1, line 2: comment here needs to be updated

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
...
+
+    set file_paths_selected ($rg_cmd $token | fzf $fzf_arguments)
+
+    if test $status -eq 0
+        # If the user was in the middle of inputting the first token and only one path is selected,
+        # then prepend ./ to the selected path so that
+        # - if the path is an executable, the user can hit Enter one more time to immediately execute it
+        # - if the path is a directory, the user can hit Enter one more time to immediately cd into it (fish will
+        #   attempt to cd implicitly if a directory name starts with a dot)
+        if test (count $file_paths_selected) = 1
```
PatrickF1, line 28: I think we can remove this logic since it'll never output a directory, and presumably we're finding source code rather than binary. I understand executable scripts are an exception but that's a very rare use case for a lot of logic. What do you think @kidonng ?

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,40 @@
...
+
+    set rg_cmd rg --ignore-case --files-with-matches
+    set redirect "< /dev/tty > /dev/tty 2>&1"
+
+    set token (commandline --current-token | string unescape)
+
+    set fzf_arguments --multi --ansi --exact --phony --query=$token \
+        --preview='rg --ignore-case --pretty --context 10 {q} {}' \
+        # reload the search when the query cange
+        --bind "change:reload:$rg_cmd {q} || true" \
```
PatrickF1, line 14: how come you have to put `|| true` at the end?

conf.d/fzf.fish
```diff
@@ -8,6 +8,7 @@ set __fzf_search_vars_cmd '__fzf_search_shell_variables (set --show | psub) (set
 if not set --query fzf_fish_custom_keybindings
     # \cf is Ctrl+f
     bind \cf '__fzf_search_current_dir'
+    bind \cg '__fzf_grep_current_dir'
```
PatrickF1, line 11: I picked g for grep. May change it later though. Open to suggestions

functions/__fzf_grep_current_dir.fish
kidonng: Makes sense.

functions/__fzf_grep_current_dir.fish
ovv: I would love to support `$EDITOR` and `$VISUAL` but the arguments are not portable. We would need a big switch statements to support various programs.

ovv:
> I love the idea of using bat though. Since bat is already a dep, we should use that to preview the occurrences of the matching string in the file.

The reason I didn't go with `bat` as preview and I added the `less` and `vim` bindings is that IMO the preview is not good enough for bigger files. Especially the bat preview might not show the match and only one if there are multiple matches.

functions/__fzf_grep_current_dir.fish
```diff
@@ -1,15 +1,14 @@
-function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
-    # Make sure that fzf uses fish to execute __fzf_preview_file.
-    # See similar comment in __fzf_search_shell_variables.fish.
+function __fzf_grep_current_dir --description "Search current directory content for current token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish.
     set --local --export SHELL (command --search fish)
 
     set rg_cmd rg --ignore-case --files-with-matches
     set redirect "< /dev/tty > /dev/tty 2>&1"
```
PatrickF1, line 6: Don't need this anymore. It was very clever though!

functions/__fzf_grep_current_dir.fish
```diff
@@ -1,15 +1,14 @@
-function __fzf_grep_current_dir --description "Search current directory content for $token. Replace the current token with the selected file paths."
-    # Make sure that fzf uses fish to execute __fzf_preview_file.
-    # See similar comment in __fzf_search_shell_variables.fish.
+function __fzf_grep_current_dir --description "Search current directory content for current token. Replace the current token with the selected file paths."
+    # Make sure that fzf uses fish.
     set --local --export SHELL (command --search fish)
 
     set rg_cmd rg --ignore-case --files-with-matches
```
PatrickF1, line 5: Love that you turned this into a variable to DRY it up!

PatrickF1:
Hey Quentin, nice work so far. Apologies for the slowness of my reviews. I've been feeling burnt out. Also, this is a major change and there are a lot of considerations I need to make regarding the key binding, the name of the function, how this is going to fit with the rest of fzf.fish, so it might be another week until we mere. Might be faster but just setting expectations, whether make each line match rather than file match an fzf result so we can use bat, how to incorporate the user's existing rg config.

In the meantime, if you really want to speed things up, you can try writing a few tests. But don't worry if you can't.

Thanks for being patient with me and contributing a great change!

PatrickF1:
Btw, shameless ask, since you've already invested so much, would you mind starring the repo? Thanks for considering it :)

ovv:
> Hey Quentin, nice work so far. Apologies for the slowness of my reviews. I've been feeling burnt out.

No worries, feel free to take all the time you need. We aren't in any hurry and your reviews are quite fast. Take care !

PatrickF1:
@ovv Tried out your new changes. Wow it works great! I wonder if we can make the highlighting more obvious though.

functions/__fzf_grep_current_dir.fish
```diff
@@ -0,0 +1,24 @@
...
+
+    set fzf_arguments --multi --ansi --delimiter=: --with-nth=1,3 --disabled --query=$token \
+        --preview 'bat --style=numbers --color=always --highlight-line {2} {1}' \
+        # center the preview window on the highlighted line
+        --preview-window +{2}-/2 \
+        # reload the search when the query change
+        --bind "change:reload:$rg_cmd {q} || true"
+
+    for file in ($rg_cmd $token | fzf $fzf_arguments)
+        set file_paths_selected $file_paths_selected (string split : $file)[1]
```
PatrickF1, line 16: have you considered doing `set --append` for better performance?
```
set file_paths_selected
for file...
   set --apend file_paths_selected (string split : $file)[1]
end
```

PatrickF1:
I'm starting to change my mind about using bat since it means
1. dumping the entire directory's contents into fzf (performance)
2. very hard to scroll through the results when there's so many results (inefficient for the user)
3. this weird UX state where users can select the same file multiple times

Sorry for the churn. Let's less go back to the non-bat version unless you or @kidonng have another idea or disagree with me (please feel free to).

ovv:
> I'm starting to change my mind about using bat since it means ...

I guess the main question is what's the workflow people usually do:

1. Are you grepping your directory to find one or more file with one or more occurrence. Then you open and edit them (the `ripgrep` preview would be good for this usecase)
2. Are you grepping your directory to find all the occurrence and preview them (checking for a function definition/usage for example)

Both sounds like valid use case to me but we can't support both right now. Since we are returning file name to the console, option 1 appears to be a better fit. OTHO I like the bat preview because you get a lot more context  on the match.

PatrickF1:
You make for a good product manager, Quentin.

I've always intended fzf.fish to primarily help put nouns into the command line. Yes, it helps with previews, but only in the name of helping the user select something to put on the command line. So in this case I think we're gonna go with 1.

ovv:
@PatrickF1 what about now ?

PatrickF1:
@ovv Nice. I will take it from here. Need to write tests, update docs, record gifs, prepare to cut a new release. Might be a few days turnaround time. Good work!

PatrickF1:
Okay new idea that's a bit more user friendly: what if the function first requests user input. And then the user can input something like `--hidden --follow -e ^something` (if they need to pass in custom opts to rg), or `something` (if they just need to do a plain search). The motivation for this idea is
- rg is very powerful and we can't predict what the user wants for their file searching needs and preset the rg opts for them. So, give users the power to pass in the opts/regex string as they see fit
- this is more performant than re-running rg every single type you run a char into the query
- frees up fzf's search functionality to filter files

Downside
- user has to re-run if they decide to change the results
- a bit slower to use, overall

I thought of this because I realized the feature as it is currently doesn't offer much more functionality than just opening up the folder in VSCode or a well-configured Emacs and just using the text editor's global search, which also comes with regex and highly configurable search.

Thoughts @ovv @kidonng? (Btw thank you both for continuing to offer feedback on this)

ovv:
Seems like a good idea but I would need to use it to make sure it fit in my usual workflow

PatrickF1:
Hi ismay. Thanks for explaining your use case. I think your use case is pretty atypical though so I'm still a no on this feature, sorry.

ismay:
Thanks for the quick reply. No worries. Though to make the case for the ubiquity of this usecase, this exact usecase is even documented in the fzf docs:

- https://github.com/junegunn/fzf#3-interactive-ripgrep-integration
- https://github.com/junegunn/fzf/blob/master/ADVANCED.md#ripgrep-integration

I feel it’s probably fair to say that it’s pretty common for it to be documented in the fzf docs. Then again, it’s up to you whether you feel it fits with the project of course, but I do feel it’s pretty common.

PR #119, opened 2021-02-06
Fix search file feature unnecessarily prepending ./

ovv:
If the file path already starts with . or /, which can happen if such a path was used as the base directory, then don't prepend ./.
Without such logic, the path becomes invalid. For example:
- if the current token is `/etc/` and you select `systemd`,  you end up with `'.//etc/systemd'`
- if the current token is `~/` and you select `Desktop`, you end up with `'./~/desktop'`

Additionally, since the output paths are quoted, we need to use the expanded token when outputting paths because ~ doesn't get expanded inside quotes.

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -23,6 +23,6 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     else if test -p "$file_path"
         __fzf_report_file_type "$file_path" "named pipe"
     else
-        echo "File doesn't exist." >&2
+        echo "File doesn't exist or is empty." >&2
     end
 end
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -5,7 +5,8 @@ function __fzf_search_current_dir --description "Search the current directory. R
 
     set fd_arguments --hidden --color=always --exclude=.git
     set fzf_arguments --multi --ansi
-    set token (commandline --current-token | string unescape)
+    set current_token (commandline --current-token)
+    set token (string unescape $current_token)
     # need to expand ~ in the directory name since fd can't expand it
     set expanded_token (string replace --regex -- "^~/" $HOME/ $token)
 
@@ -23,14 +24,16 @@ function __fzf_search_current_dir --description "Search the current directory. R
 
 
     if test $status -eq 0
-        # If the user was in the middle of inputting the first token and only one path is selected,
-        # then prepend ./ to the selected path so that
-        # - if the path is an executable, the user can hit Enter one more time to immediately execute it
-        # - if the path is a directory, the user can hit Enter one more time to immediately cd into it (fish will
-        #   attempt to cd implicitly if a directory name starts with a dot)
-        if test (count $file_paths_selected) = 1
+        # Fish will implicitly take action on a path when a path is provided as the first token and it
+        # begins with a dot, slash, or tilde. If the path is a directory, Fish will cd into it.
+        # If the path is an executable, Fish will execute it. To help users harness
+        # this convenient behavior, we automatically prepend ./ to the selected path if
+        # - only one path was selected,
+        # - the user was in the middle of inputting the first token,
+        # - and the path doesn't already begin with a dot, slash, or tilde
+        # Then, the user only needs to hit Enter once more to potentially cd into or execute that path.
+        if test (count $file_paths_selected) = 1 && not string match --regex "^[.|/|~]" $file_paths_selected
             set commandline_tokens (commandline --tokenize)
-            set current_token (commandline --current-token)
             if test "$commandline_tokens" = "$current_token"
                 set file_paths_selected ./$file_paths_selected
             end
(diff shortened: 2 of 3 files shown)


functions/__fzf_search_current_dir.fish
```diff
@@ -11,13 +11,13 @@ function __fzf_search_current_dir --description "Search the current directory. R
...
+    if string match --quiet -- "*/" $expanded_token && test -d $expanded_token
         set --append fd_arguments --base-directory=$expanded_token
         # use the directory name as fzf's prompt to indicate the search is limited to that directory
-        set --append fzf_arguments --prompt=$token --preview="__fzf_preview_file $token{}"
-        set file_paths_selected $token(fd $fd_arguments 2>/dev/null | fzf $fzf_arguments)
+        set --append fzf_arguments --prompt=$expanded_token --preview="__fzf_preview_file $expanded_token{}"
+        set file_paths_selected $expanded_token(fd $fd_arguments 2>/dev/null | fzf $fzf_arguments)
     else
-        set --append fzf_arguments --query=$token --preview='__fzf_preview_file {}'
+        set --append fzf_arguments --query=$expanded_token --preview='__fzf_preview_file {}'
```
PatrickF1, line 20: I don't see the bug. We used `token` intentionally because in all of these instances, ~ does expand to $HOME

functions/__fzf_search_current_dir.fish
```diff
@@ -29,10 +29,12 @@ function __fzf_search_current_dir --description "Search the current directory. R
...
-            set commandline_tokens (commandline --tokenize)
-            set current_token (commandline --current-token)
-            if test "$commandline_tokens" = "$current_token"
-                set file_paths_selected ./$file_paths_selected
+            if string match --invert "^/" $file_path_selected[1]
+                set commandline_tokens (commandline --tokenize)
+                set current_token (commandline --current-token)
+                if test "$commandline_tokens" = "$current_token"
+                    set file_paths_selected ./$file_paths_selected
+                end
```
PatrickF1, line 37: Can you explain the bug here, please?

functions/__fzf_search_current_dir.fish
ovv: Without those if I do a search with `~/` as token and select `desktop` I end up with `'./~/desktop'`.

The result doesn't work because it's escaped.

functions/__fzf_search_current_dir.fish
ovv: If you do a query with `/etc/` as token and select the `systemd` directory you end up with `.//etc/systemd`

PatrickF1:
Oh boy I'm starting to regret introducing the ./ feature. Thanks Quentin!

functions/__fzf_search_current_dir.fish
ovv: Did some more tests, the `expanded_token` is only needed there

PatrickF1:
Hey Quentin, let's back up. The bug you're trying to fix is basically when the current token is used as the base path and it starts with `~` or `/` right? If so, I think there's a simpler and more straightforward way to fix it. Hope you don't mind if I redo your PR a bit.

ovv:
> Hope you don't mind if I redo your PR a bit.

Sure no problem, you have a lot more experience with this code than me. Take this PR as an extended bug report :) 

> Hey Quentin, let's back up. The bug you're trying to fix is basically when the current token is used as the base path and it starts with `~` or `/` right? If so, I think there's a simpler and more straightforward way to fix it.

Exactly, I just tested your changes. They do fix the `/` case but it doesn't fix `~`.

 Any result with a `~` will be escaped by 
```
commandline --current-token --replace -- (string escape -- $file_paths_selected | string join ' ')
```
and then fish will not expand `~` to `$HOME`. We do need to keep the escaping (to handle path with a ` ` for example). The only solution I see is to replace `~` by `$HOME` ourselves.

PatrickF1:
> Exactly, I just tested your changes. They do fix the / case but it doesn't fix ~.

You're right. I caught that and was looking for a way to avoid having to always expand ~ because it makes the resulting command line kinda ugly. But it looks like I don't have a choice without terribly complicating the code. Darn

PatrickF1:
@ovv updated. Can you help me test one more time, please?

ovv:
everything looks good :shipit:

functions/__fzf_preview_file.fish
```diff
@@ -23,6 +23,6 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     else if test -p "$file_path"
         __fzf_report_file_type "$file_path" "named pipe"
     else
-        echo "File doesn't exist." >&2
+        echo "File doesn't exist or is empty." >&2
```
PatrickF1, line 26: Turns out this case is hit also when the file is empty.

kidonng:
Thank you @ovv! Yes, the tilde should be expanded because it may be quoted.

> Oh boy I'm starting to regret introducing the ./ feature.

It's certainly my oversight. I didn't notice this when making #75 and #109.

PR #120, opened 2021-02-06
Make preview directory command configurable

PatrickF1:
To acquiesce to the popularity and diversity of ls alternatives (e.g. https://github.com/ogham/exa, https://github.com/Peltoche/lsd, https://github.com/athityakumar/colorls, tree, all of which have many different options to configure them), I have finally decided to make the command used to preview directories configurable. Users can place their desired command in fzf_preview_dir_cmd and fzf.fish will use it.

Related to #117 and #101

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -3,8 +3,13 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     if test -f "$file_path" # regular file
         bat --style=numbers --color=always "$file_path"
     else if test -d "$file_path" # directory
-        set --local CLICOLOR_FORCE true
-        command ls -A "$file_path" # list hidden files as well, except for . and ..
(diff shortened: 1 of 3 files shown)


functions/__fzf_preview_file.fish
```diff
@@ -3,8 +3,13 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     if test -f "$file_path" # regular file
         bat --style=numbers --color=always "$file_path"
     else if test -d "$file_path" # directory
-        set --local CLICOLOR_FORCE true
```
PatrickF1, line 6: This wasn't working and I couldn't figure out a way to pass in `CLICOLOR_FORCE` while still using `command` to execute ls (which is necessary when people have ls aliased to something else), so I am removing it. Hopefully, passing in the additional `-F` will compensate a bit for the lack of colors.

PR #122, opened 2021-02-10
Ensure variables set by conf.d/fzf.fish are global and update uninstaller

etcusrvar:
I believe you want this to be global.
If `fisher_path` is undefined and this file is in `$__fish_config_dir/conf.d`, it will already be global by default.
If `fisher_path` is defined, and `$fisher_path/conf.d` has to be sourced manually, it will default to local, which I believe is not intended.

### DIFF
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -49,5 +49,7 @@ function _fzf_uninstall --on-event fzf_uninstall
         echo "fzf.fish key bindings removed"
         set_color normal
     end
+
+    set --erase __fzf_search_vars_cmd
     functions --erase _fzf_uninstall
 end
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -33,7 +33,7 @@ if not set --query FZF_DEFAULT_OPTS
     # height=90% leaves space to see the current command and some scrollback, maintaining context of work
     # preview-window=wrap wraps long lines in the preview window, making reading easier
     # marker=* makes the multi-select marker more distinguishable from the pointer (since both default to >)
-    set --export FZF_DEFAULT_OPTS '--cycle --layout=reverse --border --height=90% --preview-window=wrap --marker="*"'
+    set --global --export FZF_DEFAULT_OPTS '--cycle --layout=reverse --border --height=90% --preview-window=wrap --marker="*"'
 end
 
 function _fzf_uninstall --on-event fzf_uninstall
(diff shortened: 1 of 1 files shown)


PatrickF1:
Hi @etcusrvar, thanks for contributing a bug fix! I'm afraid I don't understand. What does `fisher_path` have to do with this?

etcusrvar:
I probably should have tested this a bit more before opening this issue.
TL;DR - it's a local setup issue-- but explicitly setting global may still be good idea.

In my local setup, I essentially follow [this example](https://github.com/jorgebucaran/fisher/issues/640#issue-771658021)-- but not exactly.
Instead of
```fish
for file in $fisher_path/conf.d/*.fish
    source $file
end
```
I have
```fish
function __source_fisher_confd
    for file in $fisher_path/conf.d/*.fish
        source $file
    end
end
__source_fisher_confd
```

In the first case, the default scope when sourcing conf.d/fzf.fish is global, but in the second case, being wrapped in a function, the default scope is local.

I have a few other fisher plugins installed that also set variables, but they explicitly set globals where appropriate, so I only saw this issue with your plugin. Even though the issue is with my setup, you may still want to explicitly use --global for `FZF_DEFAULT_OPTS`. (And `__fzf_search_vars_cmd`, which I can add to this PR if desired.)

PatrickF1:
Ohhh I see....I finally understand what the `set --help` docs mean:
> If  a  variable  is  not  explicitly  set  to be either universal, global or local and has never before been
  defined, the variable will be local to the currently executing function. Note that this  is  different  from
  using  the  -l  or --local flag. If one of those flags is used, the variable will be local to the most inner
  currently executing block, while without these the variable will be local to the function. If no function is
  executing, the variable will be global.

Yes, making those explicitly global is a great idea. Thanks @etcusrvar !

conf.d/fzf.fish
```diff
@@ -49,5 +49,7 @@ function _fzf_uninstall --on-event fzf_uninstall
         echo "fzf.fish key bindings removed"
         set_color normal
     end
+
+    set --erase __fzf_search_vars_cmd
```
PatrickF1, line 53: Out of scope change: while updating `__fzf_search_vars_cmd` I realized I never accounted for it in the uninstaller.

PatrickF1 (approved):

etcusrvar:
Cool. Thanks!

PR #125, opened 2021-02-14
Show command preview in history search

ovv:
Long commands get truncated in the fzf finder, while the spacing of multi-line commands are messed up. To remedy this,
show a preview of the entire command at the bottom.

### DIFF
--- a/functions/__fzf_search_history.fish
+++ b/functions/__fzf_search_history.fish
@@ -7,7 +7,7 @@ function __fzf_search_history --description "Search command history. Replace the
         fzf --read0 --tiebreak=index --query=(commandline) \
             # preview the current command in a window at the bottom 3 lines tall
             --preview="echo {4..}" \
-            --preview-window="bottom:3" |
+            --preview-window="bottom:3:wrap" |
         string collect
     )
 
--- a/functions/__fzf_search_history.fish
+++ b/functions/__fzf_search_history.fish
@@ -4,7 +4,9 @@ function __fzf_search_history --description "Search command history. Replace the
     set command_with_ts (
         # Reference https://devhints.io/strftime to understand strftime format symbols
         builtin history --null --show-time="%m-%d %H:%M:%S | " |
-        fzf --read0 --tiebreak=index --query=(commandline) |
+        fzf --read0 --tiebreak=index --query=(commandline) \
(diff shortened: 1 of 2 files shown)


functions/__fzf_search_history.fish
```diff
@@ -4,7 +4,9 @@ function __fzf_search_history --description "Search command history. Replace the
     set command_with_ts (
         # Reference https://devhints.io/strftime to understand strftime format symbols
         builtin history --null --show-time="%m-%d %H:%M:%S | " |
-        fzf --read0 --tiebreak=index --query=(commandline) |
+        fzf --read0 --tiebreak=index --query=(commandline) \
+            --preview="echo {4..}" \
+            --preview-window="bottom:10%" |
```
PatrickF1, line 9: I think you can increase this to 20%.
I'm trying to decide if locking it in at the bottom is a problem for people who customize FZF default opts to not use --reverse.

functions/__fzf_search_history.fish
```diff
@@ -4,7 +4,9 @@ function __fzf_search_history --description "Search command history. Replace the
     set command_with_ts (
         # Reference https://devhints.io/strftime to understand strftime format symbols
         builtin history --null --show-time="%m-%d %H:%M:%S | " |
-        fzf --read0 --tiebreak=index --query=(commandline) |
+        fzf --read0 --tiebreak=index --query=(commandline) \
+            --preview="echo {4..}" \
```
PatrickF1, line 8: How does `echo {4..}` result in the just the command being displayed? Wondering if there's a fish syntactical trick I can do to `set command_selected (string split --max 1 " | " $command_with_ts)[2]` below

functions/__fzf_search_history.fish
ovv: I don't see why it would be a problem for people not using `--reverse`. The preview isn't linked to the way the results are displayed.

functions/__fzf_search_history.fish
ovv: `{4..}` is an fzf trick not related to fish. fzf replace `{}` in the preview by the selected line. You can also index the line by field (by default it's split on whitespace, this can be changed with `--delimiter`). `{4..}` means from the fourth field to the end of the selected line

PR #126, opened 2021-02-15
[Search files] make fd configurable, exclude hidden files by default

PatrickF1:
Making fd configurable has been a long time ask and I'm finally giving in (e.g. #107, #112, and many other issues). It makes sense because fd has a wealth of useful options and I can see how others may want to play around with them to fit their use case.

I also have decided to exclude hidden files by default. This is a backwards break change but I doubt many people will notice. It was only ever useful for me to search my own dotfiles repo, and much of the feedback I got was asking to NOT search hidden files by default. I also made this change because it makes the new configuration code much simpler to implement.


functions/__fzf_search_current_dir.fish
```diff
@@ -3,7 +3,7 @@ function __fzf_search_current_dir --description "Search the current directory. R
     # See similar comment in __fzf_search_shell_variables.fish.
     set --local --export SHELL (command --search fish)
 
-    set fd_arguments --hidden --color=always --exclude=.git
+    set fd_opts --color=always $fzf_fd_opts
```
PatrickF1, line 6: renamed the variable because we are using opts everywhere else, e.g. FZF_DEFAULT_OPTS

tests/search_current_dir/no_base_dir_if_no_slash.fish
```diff
@@ -1,14 +1,16 @@
...
 end
 mock fzf \* ""
 mock commandline --current-token "echo functions"
 mock commandline "--current-token --replace" ""
 mock commandline \* ""
 __fzf_search_current_dir
+
 if test -d functions
-    @test "doesn't change fd's base directory if no slash on current token" -z (string match --entire -- "--base-directory" $fd_args)
+    test -n "$fd_captured_opts" && test -z (string match --entire -- "--base-directory" $fd_captured_opts)
```
PatrickF1, line 12: Out of scope: this test wasn't failing even when the code wasn't working.

PR #127, opened 2021-02-16
Check markdown and yaml format in CI

PatrickF1:


PatrickF1:
Note to self: doesn't seem to be working for yaml.

PR #129, opened 2021-03-01
Check Fish syntax and format in CI

IlanCosman:
I recently created two new github actions for Fish, [format-check](https://github.com/fish-actions/format-check), and [syntax-check](https://github.com/fish-actions/syntax-check). This PR adds them to the CI workflow. I also ran Fish 3.2's new `fish_indent` on the codebase so that `format-check` would be happy.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -32,6 +32,26 @@ jobs:
         run: fishtape tests/*/*
         shell: fish {0}
 
+  syntax-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: fish-actions/install-fish@v1
+
+      - uses: fish-actions/syntax-check@v1
+
+  format-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: fish-actions/install-fish@v1
+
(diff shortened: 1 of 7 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -32,6 +32,26 @@ jobs:
         run: fishtape tests/*/*
         shell: fish {0}
 
+  syntax-check:
```
PatrickF1, line 35: I know this is a bit out-of-scope, but is there any significance to the order the jobs are defined and run? I'm curious if we'd want the syntax check to run first before the tests, because obviously if the syntax check fails, the tests will as well. But I forget if GitHub actions will stop running the rest of the tests if one fails, or if the order defined = order run. Do you know?

PatrickF1 (approved):
Hey @IlanCosman, wow, thanks for creating those Fish GitHub Actions and helping me incorporate them! They are going to come in really handy :)

.github/workflows/continuous_integration.yml
IlanCosman: No. In a matrix, one job fail will cause the others to cancel. But otherwise, no, jobs do not effect each other. 

Source: https://github.com/IlanCosman/tide/actions/runs/604243138

PatrickF1: Good to know, thanks!

jorgebucaran (approved):

PR #133, opened 2021-03-09
[Search directory] expand variables in current token

kidonng:
Evolved from #109. Now not only does tilde (~) in the current token get expanded, variables do as well. This is useful for quickly accessing known directory variables, e.g. $XDG_CONFIG_HOME, $TMPDIR, $JAVA_HOME, $fisher_path, $__fish_user_data_dir.

image

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -1,11 +1,10 @@
 # helper function for __fzf_search_current_dir
-function __fzf_preview_file --description "Print a preview for the given file based on its file type."
-    set file_path $argv
+function __fzf_preview_file --argument-names file_path --description "Print a preview for the given file based on its file type."
     if test -f "$file_path" # regular file
         bat --style=numbers --color=always "$file_path"
     else if test -d "$file_path" # directory
         if set --query fzf_preview_dir_cmd
-            command $fzf_preview_dir_cmd "$file_path"
+            eval $fzf_preview_dir_cmd \"$file_path\"
         else
             # -A list hidden files as well, except for . and ..
             # -F helps classify files by appending symbols after the file name
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -28,6 +28,6 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     else if test -p "$file_path"
         __fzf_report_file_type "$file_path" "named pipe"
     else
-        echo "File doesn't exist or is empty." >&2
+        echo "$file_path doesn't exist." >&2
     end
 end
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -1,5 +1,6 @@
 # helper function for __fzf_search_current_dir
-function __fzf_preview_file --argument-names file_path --description "Print a preview for the given file based on its file type."
+function __fzf_preview_file --description "Print a preview for the given file based on its file type."
+    set file_path $argv
     if test -f "$file_path" # regular file
         bat --style=numbers --color=always "$file_path"
     else if test -d "$file_path" # directory
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -4,7 +4,7 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
         bat --style=numbers --color=always "$file_path"
     else if test -d "$file_path" # directory
         if set --query fzf_preview_dir_cmd
-            eval $fzf_preview_dir_cmd "$file_path"
+            command $fzf_preview_dir_cmd "$file_path"
         else
             # -A list hidden files as well, except for . and ..
             # -F helps classify files by appending symbols after the file name
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -4,6 +4,7 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
         bat --style=numbers --color=always "$file_path"
     else if test -d "$file_path" # directory
         if set --query fzf_preview_dir_cmd
+            # need to escape quotes to make sure eval receives file_path as a single arg
             eval $fzf_preview_dir_cmd \"$file_path\"
         else
             # -A list hidden files as well, except for . and ..
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -5,18 +5,19 @@ function __fzf_search_current_dir --description "Search the current directory. R
 
     set fd_opts --color=always $fzf_fd_opts
     set fzf_arguments --multi --ansi
-    set current_token (commandline --current-token)
-    set token (string unescape -- $current_token)
-    # expand the token (which may include tilde, variables, etc.) into full path
+ set token (commandline --current-token)
+    # expand ~ in the directory name since fd can't expand it
     set expanded_token (eval echo -- \"$token\")
+    # unescape token because it's already quoted so backslashes will mess up the path
+    set unescaped_exp_token (string unescape -- $expanded_token)
 
     # If the current token is a directory and has a trailing slash,
     # then use it as fd's base directory.
     if string match --quiet -- "*/" $token && test -d "$unescaped_exp_token"
         set --append fd_opts --base-directory=$unescaped_exp_token
         # use the directory name as fzf's prompt to indicate the search is limited to that directory
-        set --append fzf_arguments --prompt=$token --preview="__fzf_preview_file $expanded_token{}"
-        set file_paths_selected $expanded_token(fd $fd_opts 2>/dev/null | fzf $fzf_arguments)
+        set --append fzf_arguments --prompt="$unescaped_exp_token" --preview="__fzf_preview_file $expanded_token{}"
+        set file_paths_selected $unescaped_exp_token(fd $fd_opts 2>/dev/null | fzf $fzf_arguments)
     else
         set --append fzf_arguments --query="$unescaped_exp_token" --preview='__fzf_preview_file {}'
         set file_paths_selected (fd $fd_opts 2>/dev/null | fzf $fzf_arguments)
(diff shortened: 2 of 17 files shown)


kidonng:
In addition to the intended changes of this PR (which is just two lines), I fixed two other issues:

1. `__fzf_preview_file` may receive a path containing spaces, so we should directly use `$argv` as filename.
2. Even if we fixed 1, `eval` will wreck the path, even if you escape it.

Funny how this feature also makes use of `eval` 😂

But changing `eval` to `command` `has its own issues:

- Test breaks (fixed)
- Users can't specify a function anymore. Though I doubt if there are  really needs for this.

kidonng:
I wonder how will this break things (the fixed bugs is not related to this change). It's literally just echo strings.

tests/preview_file/custom_dir_preview.fish
```diff
@@ -1,4 +1,5 @@
-set fzf_preview_dir_cmd exa
-mock exa . "echo file.txt"
-set actual (__fzf_preview_file .)
+set fzf_preview_dir_cmd ls
+set dir (mktemp -d)
+touch $dir/{file.txt,.hidden}
```
PatrickF1, line 3: Why did you have to change this test?

kidonng: We are using command instead of eval now, but `mock` only creates functions, so the test failed: https://github.com/PatrickF1/fzf.fish/commit/e24052f1b56db82658bf78a82fb7c96f5f78f28e

functions/__fzf_preview_file.fish
```diff
@@ -1,10 +1,11 @@
 # helper function for __fzf_search_current_dir
-function __fzf_preview_file --argument-names file_path --description "Print a preview for the given file based on its file type."
+function __fzf_preview_file --description "Print a preview for the given file based on its file type."
+    set file_path $argv
```
PatrickF1, line 3: I can't repro the bug that this fixes. I created a dir named "dir with space" and put some files, also containing spaces in the name, and was able to preview them just fine using the code on `main`. Can you teach me how to repro the bug?

tests/preview_file/custom_dir_preview.fish
PatrickF1: gotcha, thanks for explaining!

PatrickF1:
@kidonng Mind I'm ready to merge in the main change but not the bug fix. Mind if I or you revert the unnecessary changes and merge them in later?

functions/__fzf_preview_file.fish
PatrickF1: Ah ok I see why this bug happens. It was introduced by this PR, because we have to pass the unescaped expanded token to `__fzf_preview_file`. This is a good example of additional features that introduces bugs and more complicated code...

kidonng: Sorry for the delay!

> Can you teach me how to repro the bug?

Actually I just realized there's a implicit prerequisite: you have to set `fzf_preview_dir_cmd` or the code won't run `eval $fzf_preview_dir_cmd "$file_path"` but just regular `command ls -A -F "$file_path"`.

That said, here is a video demo of the issue:

https://user-images.githubusercontent.com/44045911/111025553-d0fc9980-841f-11eb-9bb5-fd6dd646fa00.mov

> It was introduced by this PR

No it's not. We do need to pass `expanded_token` now, but that's not how the bug works.

kidonng:
> Mind I'm ready to merge in the main change but not the bug fix. Mind if I or you revert the unnecessary changes and merge them in later?

That's ok if you want to, but I'm not in favor of the idea because:

- This feature is affected by the bug
- The new PR still needs to refer this one
- The fix is not a lot of code (unless you have to include the test changes)

PatrickF1:
> That's ok if you want to, but I'm not in favor of the idea because:
> 
> * This feature is affected by the bug
> * The new PR still needs to refer this one
> * The fix is only 3 LOC (unless you have to include the test changes)

True true we'll keep it in this PR. Let me reset it for you.
I don't know what you meant by 
> Actually I just realized there's a implicit prerequisite: you have to set fzf_preview_dir_cmd or the code won't run eval $fzf_preview_dir_cmd "$file_path" but just regular command ls -A -F "$file_path".

Can you fix that by taking over this PR again?

kidonng:
> Can you fix that by taking over this PR again?

Perhaps I'm not expressing myself clearly, I was saying that probably you didn't set `fzf_preview_dir_cmd` so can't reproduce.

Maybe `eval` can be kept, we just need to quote the arguments properly.

functions/__fzf_search_current_dir.fish
```diff
@@ -7,15 +7,15 @@ function __fzf_search_current_dir --description "Search the current directory. R
...
+    # expand the token (which may include tilde, variables, etc.) into full path
+    set expanded_token (eval echo -- \"$token\")
 
     # If the current token is a directory and has a trailing slash,
     # then use it as fd's base directory.
     if string match --quiet -- "*/" $token && test -d $expanded_token
         set --append fd_opts --base-directory=$expanded_token
         # use the directory name as fzf's prompt to indicate the search is limited to that directory
-        set --append fzf_arguments --prompt=$token --preview="__fzf_preview_file $token{}"
+        set --append fzf_arguments --prompt=$token --preview="__fzf_preview_file $expanded_token{}"
```
PatrickF1, line 18: Interesting solution adding escaped quotes around the token. What if you just added single quotes here
```
--preview="__fzf_preview_file '$expanded_token{}'"
```
?

PatrickF1: oh nvm that wouldn't work because single quotes prevent string interpolation...hmmm okay

PatrickF1: TBH it seems to work but I don't like the fix. It's too hacky, hard to read, hard to understand. Let me see if I can fix it myself. I started a new branch because it's a complicated fix and I realized I'm going to have to add tons more test and documentation.

PatrickF1: Gonna work on it here https://github.com/PatrickF1/fzf.fish/pull/137

kidonng: > It's too hacky, hard to read, hard to understand.

The fix is only adding two backslashes in __fzf_preview_file, how in the world is it complicated? And the problematic eval isn't even introduced in this PR.

As for the feature itself only two other lines are changed.

I think we will never get there since I don't see any simper/better solution. If you really find this expanding feature troublesome, let's just skip it.

PatrickF1: > The fix is only adding two backslashes in __fzf_preview_file, how in the world is it complicated?

It is complicated: I've never seen any shell code force quotes except when needing to put quotes inside a string for printing, not running a command. Maybe I'm obtuse, but the multiple levels of interpolation and evaluation of strings and commands and arguments in this file is mind boggling. It's hard to grasp, much harder to document and test.

> As for the feature itself only two other lines are changed.

True, that's why I'm willing to merge it. I'll merge it as soon as I figure out a better fix that's hopefully simpler.

> If you really find this expanding feature troublesome, let's just skip it.

To be perfectly transparent so there's no surprises, I get disturbed just thinking about how complicated `__fzf_search_current_dir.fish` is. I really, really want to cut back on the number of features in it. And the most complicated one of them of all is the expansion of ~ and $vars. I am going to try to keep it because we've already put in the work and I know you and some other people really want it and use it. But personally, it's eating up what little time I have to devote to this project so ultimately it hurts everyone. If I can't find a simple fix for this, I might just remove the feature. Sorry.

kidonng: > I've never seen any shell code force quotes except when needing to put quotes inside a string for printing, not running a command.

Fine, let me explain it *clearly*.
`eval` takes arguments right? But if you just quote the strings directly, **the eval sees them as regular arguments without quotes.** This is not related to eval, not even Fish, it's just how passing arguments works.
With escaped quotes, you tell the shell treat the quotes as part of the string, so the eval executes the code as expected.

PatrickF1:
@kidonng Finally merged #137. Can you fix the merge conflicts and we may finally merge this today? Thanks! Sorry for the long wait.

kidonng:
Should be covered by existing test for tilde expanding but we can also add tests for variable expanding if you want.

PatrickF1:
Yes please, could you add a test for the TMPDIR? An idea I have is to execute mktmp and then check for it in TMPDIR. Up to you.

functions/__fzf_search_current_dir.fish
```diff
@@ -6,8 +6,8 @@ function __fzf_search_current_dir --description "Search the current directory. R
     set fd_opts --color=always $fzf_fd_opts
     set fzf_arguments --multi --ansi $fzf_dir_opts
     set token (commandline --current-token)
-    # expand ~ in the directory name since fd can't expand it
-    set expanded_token (string replace --regex -- "^~/" $HOME/ $token)
+    ​# expand the token (which may include tilde, variables, etc.) into full path
+    set expanded_token (eval echo -- \"$token\")
```
PatrickF1, line 10: What's an example where removing the quotes would break this?

kidonng: I'm just quoting it like in `__fzf_preview_file`.

kidonng:
Nice work there 👍 sorry for always letting you finishing the tests.

functions/__fzf_search_current_dir.fish
PatrickF1: I just can't think of when token when ever be separated by an unescaped space. Whatever, I suppose it's safer this way.

PatrickF1:
> Nice work there 👍 sorry for always letting you finishing the tests.

No problem, it's teamwork! You've also brought a lot of good ideas and expanded my perspective on what a good plugin is.

functions/__fzf_search_current_dir.fish
kidonng: [You](https://github.com/PatrickF1/fzf.fish/pull/137#issuecomment-801632385) are totally right... I was too careful to do it right.

PatrickF1: Oh shoot, this is actually a bug. Quoting it broke this for tilde. Just tested.

PatrickF1: Fixed in https://github.com/PatrickF1/fzf.fish/commit/83cb78fbb27dd969d20daace826de9c78216067d

PR #134, opened 2021-03-09
[Search history] display command with bat in preview window

kidonng:
I didn't realize I need a preview window until you added it. Nice work!
But I'm shocked (😄) to find you missed the opportunity to make it sexier.

image

### DIFF
--- a/functions/__fzf_search_history.fish
+++ b/functions/__fzf_search_history.fish
@@ -6,7 +6,7 @@ function __fzf_search_history --description "Search command history. Replace the
         builtin history --null --show-time="%m-%d %H:%M:%S | " |
         fzf --read0 --tiebreak=index --query=(commandline) \
             # preview current command in a window at the bottom 3 lines tall
-            --preview="printf {4..} | bat --plain --color=always --language=fish" \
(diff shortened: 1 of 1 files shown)


PatrickF1 (approved):
Oh my gosh what a great idea! What would this plugin do without you? Thank you for another awesome PR!

functions/__fzf_search_history.fish
```diff
@@ -6,7 +6,7 @@ function __fzf_search_history --description "Search command history. Replace the
         builtin history --null --show-time="%m-%d %H:%M:%S | " |
         fzf --read0 --tiebreak=index --query=(commandline) \
             # preview current command in a window at the bottom 3 lines tall
-            --preview="echo {4..}" \
+            --preview="echo -- {4..} | bat --plain --color=always --language=fish" \
```
PatrickF1, line 9: I noticed --plain disables pager but then I tested it anyway and the pager is still used for long commands. What gives? 🤔

kidonng: `--plain` doesn't disable pager, AFAICT.

PR #135, opened 2021-03-09
[Search history] preview commands with fish_indent --ansi instead of bat

IlanCosman:
I happened to be visiting this repository, and I found a use for the trick I learned a little bit ago! Great idea @kidonng.

Sidenote, at this point I would argue for doing
```fish
# Make sure that fzf uses fish to execute __fzf_preview_file.
```
in every command by default somehow. We're already doing it in the majority.

### DIFF
--- a/functions/__fzf_search_history.fish
+++ b/functions/__fzf_search_history.fish
@@ -2,7 +2,7 @@ function __fzf_search_history --description "Search command history. Replace the
     # history merge incorporates history changes from other fish sessions
     builtin history merge
 
-    # Make sure that fzf uses fish to execute __fzf_preview_file.
+    # Make sure that fzf uses fish so we can run fish_indent.
     # See similar comment in __fzf_search_shell_variables.fish.
     set --local --export SHELL (command --search fish)
 
--- a/functions/__fzf_search_history.fish
+++ b/functions/__fzf_search_history.fish
@@ -1,12 +1,17 @@
 function __fzf_search_history --description "Search command history. Replace the command line with the selected command."
     # history merge incorporates history changes from other fish sessions
     builtin history merge
+
+    # Make sure that fzf uses fish to execute __fzf_preview_file.
+    # See similar comment in __fzf_search_shell_variables.fish.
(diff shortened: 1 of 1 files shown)


kidonng:
Awesome! You know what? I was searching for a syntax highlighter when making that PR and suddenly remembered bat. I did saw this a few times, but I can't remember.

PatrickF1:
Any reason to prefer fish_ident over bat besides that fish_ident ships with fish?

> Sidenote, at this point I would argue for doing [...] in every command by default somehow.

Yeah it's on my todo list. I just haven't had much free time to work on it as you can tell with the other two pending PRs.

IlanCosman:
Yes, the colors look much better. In fact, the colors look exactly like how the Fish commandline would highlight the text, so that's an added benefit.

PatrickF1 (approved):
Sold! Thank you @IlanCosman

PatrickF1:
@IlanCosman I just realized fish_indent is installed as a binary in PATH.
```
which fish_indent
/usr/local/bin/fish_indent
```
so setting SHELL=fish was actually unnecessary.

kidonng:
It doesn't hurt, and we may need it in the future.

IlanCosman:
Huh, I didn't realize that. I didn't even consider that it could be anything but a builtin 😄@faho could you shed some light?

faho:
`fish_indent` is just a normal binary, the shell should have no bearing on it.

The comment in __fzf_search_shell_variables.fish is about executing a *fish function* - in that case it *is* necessary because you need fzf to tell fish to run it.

IlanCosman:
Yeah, I get that. I'm just wondering why it is that way. What's different about `fish_indent` when compared to e.g `realpath` or any other random builtin?

faho:
It's *not* a builtin. It's an external command, so you can also e.g. call it from an editor, or as part of a pygments highlighter (this is what fish does for its docs).

IlanCosman:
Got it, thanks for explaining that Fabian 😄

PR #137, opened 2021-03-14
[Search files] fix preview for paths containing spaces

PatrickF1:
@kidonng and I separately found two different bugs around paths containing spaces. Boy this search files code is getting complicated!

# Bug 1
This bug happens when the current token is a directory and fzf.fish uses it as the base directory to search for files in.
e.g. 
1. create dir named "folder with spaces"
2. add a file in that folder with some text
3. type in "folder with spaces/" and with the cursor on that token, execute search files
4. notice the preview function doesn't work on the file in the folder

# Bug 2
This bug was found by @kidonng. When using a custom preview directory command (`$fzf_preview_dir_cmd`) is set, this line from `__fzf_preview_file.fish`

```
eval $fzf_preview_dir_cmd "$file_path"
```
results in this code being executed:
```
$fzf_preview_dir_cmd $file_path
``` 
Notice there are no quotes around `file_path`, which means if `file_path` is more than one word, it gets broken up into multiple args.
The solution is to escape the quotes around `file_path` to make sure eval understands it is to be passed as a single arg to `$fzf_preview_dir_cmd`.


kidonng:
If you don't unescape the token first, the expand may fail. For example `'~/foo'`.

Again I'm not seeing any other way around, please read my comment in https://github.com/PatrickF1/fzf.fish/pull/133#discussion_r593990674.

PatrickF1:
> But if you don't unescape the token first, the expand may fail. For example it doesn't match '~/foo'

Is that the only problem you see with this PR? Cause I don't mind not supporting that. When would the user by typing in '~/foo` and expecting the ~ to be expanded. Most users would know that tildes don't expand inside a string so why expect fzf.fish to expand it?

PatrickF1:
@kidonng yeah I still can't figure out why a user would have their current token be a string containing ~. Autocomplete wouldn't quote it and users should know better than to quote ~. I'm going to merge this unless I hear from you tomorrow.

kidonng:
Please ignore that bad example.

To be clear, I'm OK with these changes, but it doesn't fix the issue.

To quote [my self](https://github.com/PatrickF1/fzf.fish/pull/133#discussion_r593990674):

> eval takes arguments right? But if you just quote the strings directly, the eval sees them as regular arguments without quotes.
> This is not related to eval, not even Fish, it's just how passing arguments works. Do you see quotes in the output of echo 'hello'?
> With escaped quotes, you tell the shell to treat the quotes as part of the string, so the eval works as expected.

PatrickF1:
Thanks for explaining it the quotes and the all the interpolation and escaping logic going on in fzf_seatch_current_dir. It makes sense to me now. I agree that this change you're making now is the easiest way to fix the bug _in its current state_. However, that's not to say we can't do some refactoring to make the code more readable as it is. Which is why I'd still like to merge this.

kidonng:
Then please include the escapes in this PR or else it doesn't match the title (and description).

PatrickF1:
> Then please include the escapes in this PR or else it won't match the title (and description).

Done. Thanks. I just realized we were miscommunicating for a long long time because there were actually 2 different bugs and I was talking about one (__fzf_preview_file receiving unescaped token) and you were talking about the other (eval receiving file_path as multiple args).

kidonng:
> __fzf_preview_file receiving unescaped token
As far as I can see there's no issue with it.

PatrickF1:
> As far as I can see there's no issue with it (and whether you change it or not doesn't matter).

Did you try to repro it using the 4 steps I outlined above in my first post?

kidonng:
I see, I fixed this in my PR previously (by using `$argv` directly, which is now reverted), so I thought they were the same thing.

PatrickF1:
> I see, I fixed this in my PR previously (by using $argv directly, which is now reverted), so I thought they were the same thing.

Oh yes, that was the other thing that made our communication difficult! I thought your original solution was very complicated and that's why I said "the multiple levels of interpolation and evaluation of strings and commands and arguments in this file is mind boggling". Man it's quite hard to communicate on GitHub. Anyway, appreciate your patience and our cooperation on this.

kidonng:
Ah, wait, I did realize they are two different bugs: https://github.com/PatrickF1/fzf.fish/pull/133#issuecomment-793803559.

And I kinda got confused by the whole discussion

PatrickF1:
Oh you're right, I think I must've had a hard time understanding it, glossed over it, and assumed it was just one bug.

kidonng:
Well, though laborious, these discussions are fruitful:

- `eval` is preserved since we are solving it properly
- Using escaped token is more explicit than using `$argv`
- Refactoring make code more readable

All things should be clear know, we should focus on gettings the bugs fixed and feature delivered.

functions/__fzf_preview_file.fish
```diff
@@ -28,6 +29,6 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     else if test -p "$file_path"
         __fzf_report_file_type "$file_path" "named pipe"
     else
-        echo "File doesn't exist or is empty." >&2
+        echo "$file_path doesn't exist or is empty." >&2
```
PatrickF1, line 32: Outputting file_path because it helps with debugging.

functions/__fzf_preview_file.fish
```diff
@@ -4,7 +4,8 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
         bat --style=numbers --color=always "$file_path"
     else if test -d "$file_path" # directory
         if set --query fzf_preview_dir_cmd
-            eval $fzf_preview_dir_cmd "$file_path"
+            # need to escape quotes to make sure eval receives file_path as a single arg
+            eval $fzf_preview_dir_cmd \"$file_path\"
```
PatrickF1, line 8: This is the fix for bug 2.

functions/__fzf_preview_file.fish
```diff
@@ -28,6 +29,6 @@ function __fzf_preview_file --argument-names file_path --description "Print a pr
     else if test -p "$file_path"
         __fzf_report_file_type "$file_path" "named pipe"
     else
-        echo "File doesn't exist or is empty." >&2
+        echo "$file_path doesn't exist." >&2
```
PatrickF1, line 32: Outputting file_path because it helps with debugging.
Removing the "is empty" part because empty files count as regular files and so will result in an empty preview.

functions/__fzf_search_current_dir.fish
```diff
@@ -5,20 +5,20 @@ function __fzf_search_current_dir --description "Search the current directory. R
...
     # If the current token is a directory and has a trailing slash,
     # then use it as fd's base directory.
-    if string match --quiet -- "*/" $token && test -d $expanded_token
-        set --append fd_opts --base-directory=$expanded_token
+    if string match --quiet -- "*/" $token && test -d "$unescaped_token"
+        set --append fd_opts --base-directory=$unescaped_token
         # use the directory name as fzf's prompt to indicate the search is limited to that directory
-        set --append fzf_arguments --prompt=$token --preview="__fzf_preview_file $token{}"
-        set file_paths_selected $expanded_token(fd $fd_opts 2>/dev/null | fzf $fzf_arguments)
+        set --append fzf_arguments --prompt="$unescaped_token" --preview="__fzf_preview_file $expanded_token{}"
```
PatrickF1, line 18: Here's bugfix 2: we need to pass the escaped (but expanded) token to __fzf_preview_file in order for the preview to work for paths containing spaces.

functions/__fzf_search_current_dir.fish
```diff
@@ -5,20 +5,21 @@ function __fzf_search_current_dir --description "Search the current directory. R
...
     # If the current token is a directory and has a trailing slash,
     # then use it as fd's base directory.
-    if string match --quiet -- "*/" $token && test -d $expanded_token
-        set --append fd_opts --base-directory=$expanded_token
+    if string match --quiet -- "*/" $token && test -d "$unescaped_exp_token"
+        set --append fd_opts --base-directory=$unescaped_exp_token
         # use the directory name as fzf's prompt to indicate the search is limited to that directory
-        set --append fzf_arguments --prompt=$token --preview="__fzf_preview_file $token{}"
-        set file_paths_selected $expanded_token(fd $fd_opts 2>/dev/null | fzf $fzf_arguments)
+        set --append fzf_arguments --prompt="$unescaped_exp_token" --preview="__fzf_preview_file $expanded_token{}"
```
PatrickF1, line 19: Here's bugfix 2: we need to pass the expanded but not yet unescaped token to `__fzf_preview_file` in order for the preview to work for paths containing spaces.

tests/preview_file/custom_dir_preview.fish
```diff
@@ -1,4 +1,6 @@
-set fzf_preview_dir_cmd exa
-mock exa . "echo file.txt"
-set actual (__fzf_preview_file .)
-@test "correctly uses custom command to preview directories" "$actual" = file.txt
+set multi_word_dir "tests/_resources/multi word dir"
+
+set fzf_preview_dir_cmd ls
+
+set actual (__fzf_preview_file "$multi_word_dir")
+@test "correctly uses custom command to preview directories" "$actual" = "file 1.txt file 2.txt"
```
PatrickF1, line 6: This test protects against a regression of bug 2

tests/search_current_dir/base_dir_with_spaces.fish
```diff
@@ -0,0 +1,8 @@
+mock commandline --current-token "echo tests/_resources/multi word dir/"
+mock commandline "--current-token --replace --" "echo \$argv"
+mock commandline \* ""
+set --export --append FZF_DEFAULT_OPTS "--filter=''" # automatically select all input lines sent to fzf
+
+set actual (__fzf_search_current_dir)
+set expected "'tests/_resources/multi word dir/file 1.txt' 'tests/_resources/multi word dir/file 2.txt'"
+@test "uses current token as base directory if it ends in / and is a directory" "$actual" = $expected
```
PatrickF1, line 8: This test protects against a regression of bug 1

.github/workflows/continuous_integration.yml
```diff
@@ -28,8 +28,15 @@ jobs:
           git clone --depth 1 https://github.com/junegunn/fzf.git ~/.fzf
           ~/.fzf/install
 
+      - name: Install fd
```
PatrickF1, line 31: Had to add installation of fd because one of the new tests had to make full of it

kidonng:
Good work 👍

.github/workflows/continuous_integration.yml
kidonng: Actually you can use Homebrew on both macOS and Linux in GitHub Actions, so these can be simplified into `brew install fzf fd`

functions/__fzf_search_current_dir.fish
kidonng: Not a fan of using expanded token as prompt since it looks bloated... But given we will be adding variable expanding soon and it's better explicit than explicit, I will pass.

PatrickF1: Yeah I actually had your PR in mind when I changed it.

tests/search_current_dir/base_dir_with_spaces.fish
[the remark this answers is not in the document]
kidonng: Just realized we never really fixed bug 1.

image

functions/__fzf_search_current_dir.fish
kidonng, line 19: The filename should have been quote here:

```suggestion
        set --append fzf_arguments --prompt="$unescaped_exp_token" --preview="__fzf_preview_file '$expanded_token{}'"
```

tests/search_current_dir/base_dir_with_spaces.fish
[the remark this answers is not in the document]
PatrickF1: How would you modify the test to catch this new case? I don't understand why it wasn't caught.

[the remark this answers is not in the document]
PatrickF1: I'm unable to repro this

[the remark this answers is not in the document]
kidonng: > How would you modify the test to catch this new case? I don't understand why it wasn't caught.

Not to my knowledge, sorry. May take a look when I get the time. 

> I'm unable to repro this

I can confirm it's not related to my shell profile. Cam you reproduce with `mkdir 'with spaces'`, type `./with\ spaces` and trigger the function?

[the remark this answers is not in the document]
PatrickF1: Nvm I see that the test doesn't test the preview function, which I don't know how to do.

I can repro now. I made a mistake before. Thanks for reporting this!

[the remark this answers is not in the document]
PatrickF1: Crazy I made the PR about fixing this bug but I didn't actually fix it 🤦‍♂️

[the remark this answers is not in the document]
kidonng: > Crazy I made the PR about fixing this bug but I didn't actually fix it 🤦‍♂️

Well I was quite confused I didn't notice it either back then.

A good lesson on always testing changes manually if possible 😄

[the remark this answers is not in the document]
kidonng: > Nvm I see that the test doesn't test the preview function, which I don't know how to do.

Maybe replace the preview function _file_ with a test function, which tests if the command output is expected.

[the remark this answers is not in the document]
PatrickF1: > Maybe replace the preview function file with a test function, which tests if the command output is expected.

I can't get fzf to execute the search function in scripted mode. It'd have to not exit automatically, but if it doesn't exit automatically, how do I make it exit so that the test can continue to run.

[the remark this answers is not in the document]
kidonng: Start a background shell, sleep for a few seconds and kill all `fzf` processes?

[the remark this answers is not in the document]
PatrickF1: Thanks for the idea. I might use it if depending on what happens to this issue https://github.com/junegunn/fzf/issues/2442.

kidonng:
Post mortem:
- Bug 1 is introduced in #75, caused by `__fzf_preview_file` receiving unescaped arguments. It is fixed in #137 by leaving the arguments as is.
- Bug 2 is introduced in #120, caused by `eval` messing around with arguments. There is two attempts to fix it: #137 and #152, but it is not completely fixed.

PR #138, opened 2021-03-17
Add git diff preview to Modified paths function

kidonng:
Inspired by [forgit](https://github.com/wfxr/forgit):

image

In fact this is the only function without a preview window now.

### DIFF
--- a/functions/__fzf_git_diff.fish
+++ b/functions/__fzf_git_diff.fish
@@ -1,5 +1,5 @@
 function __fzf_git_diff --argument-names file_status file_path --description "Print git diff for the given file based on its status."
-    if test $file_status = "A" || test $file_status = "??"
+    if test $file_status = A || test $file_status = "??"
         git diff --color=always --no-index -- /dev/null $file_path
     else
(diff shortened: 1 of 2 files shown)


PatrickF1:
Hmm there was a similar PR opened not too long ago https://github.com/PatrickF1/fzf.fish/pull/124. My issue with it was that the user can't tell whether the diff is staging vs HEAD or worktree vs HEAD. What do you think?

kidonng:
Thanks for the pointer. I didn't read that PR throughoutly, but this PR is the same as what you get with regular `git diff`, which should not be confusing.

PatrickF1:
Here's the main thing to read from that PR
Screen Shot 2021-03-22 at 10 46 34 AM

PatrickF1:
See https://github.com/PatrickF1/fzf.fish/pull/124#issuecomment-812622468. LMK if you'd want to work on it. I'm currently biased to letting Quentin finish his work since he came up with the idea first and his PR is more polished.

PatrickF1:
@kidonng A similar PR, #237, has been merged!

PR #141, opened 2021-03-23
Use Homebrew to install fzf and fd on CI

kidonng:
Who knew https://docs.brew.sh/Homebrew-on-Linux existed? This is at least 5x faster and way less code.
Follow up from https://github.com/PatrickF1/fzf.fish/pull/137/files#r599225253

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -23,17 +23,8 @@ jobs:
         with:
           plugins: jorgebucaran/fishtape ilancosman/clownfish $GITHUB_WORKSPACE
 
-      - name: Install fzf
-        run: |
-          git clone --depth 1 https://github.com/junegunn/fzf.git ~/.fzf
-          ~/.fzf/install
-
-      - name: Install fd
-        run: |
-          git clone https://github.com/sharkdp/fd
-          cd fd
-          cargo build
-          cargo install --path .
+      - name: Install fzf and fd
+        run: brew install fzf fd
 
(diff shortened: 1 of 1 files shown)


PatrickF1:
Haha thanks! Nice try but that won't work for Linux, which is necessary because I want CI to run tests on both Linux and Mac.

PatrickF1:
For fd, I have on my todo list to look into https://github.com/pablopunk/fd-find. No problem if you can't, but it would be super helpful if you could make that work.

kidonng:
Hmm you don't know Linuxbrew? And the tests did passed.

PatrickF1:
Holy smokes you're right--I just noticed they passed. That is amazing they have the same interface! Okay I'm super excited to merge this. It's almost 10 times faster and way less code!

PatrickF1 (approved):
This is so much better. Really appreciate it :)

PatrickF1:
I wonder if this would work for installing fish as well 🤔 
Probably should have tried it before merging.

kidonng:
I was going to answer that. Technically we could, but since Fish is the most important component of the test, I prefer it to be handle by the corresponding package manager of each OS, which may provide additional patch/benefits targeted for each OS.

kidonng:
Sidenote: I was also surprised about how well Linuxbrew works when I found out it. They are super useful for tests intended for both macOS and Linux, since Snap/Flatpak/AppImage doesn't exist on macOS and they are less up-to-date and has troublesome sandboxing/etc.

PR #142, opened 2021-03-23
Remove problematic character from __fzf_search_current_dir

fezboy:
There was an zero-width space hiding before a comment throwing an error. (the character's invisible in a browser or terminal, but running `git diff dcf533 2c0fb5 | vim -` should make it visible)

### DIFF
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -6,7 +6,7 @@ function __fzf_search_current_dir --description "Search the current directory. R
     set fd_opts --color=always $fzf_fd_opts
     set fzf_arguments --multi --ansi $fzf_dir_opts
     set token (commandline --current-token)
-    # expand the token (which may include tilde, variables, etc.) into full path
+    # expand any variables or leading tilde (~) in the token
     set expanded_token (eval echo -- \"$token\")
     # unescape token because it's already quoted so backslashes will mess up the path
     set unescaped_exp_token (string unescape -- $expanded_token)
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -6,7 +6,7 @@ function __fzf_search_current_dir --description "Search the current directory. R
     set fd_opts --color=always $fzf_fd_opts
     set fzf_arguments --multi --ansi $fzf_dir_opts
     set token (commandline --current-token)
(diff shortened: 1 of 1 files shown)


PatrickF1:
Wow so interesting! What is the error exactly and when does it trigger? When you run search directory? Or when you open up the file?
Btw `git diff dcf533 2c0fb5 | vim -` doesn't work for me because I don't have dcf533 anywhere.

cc: @kidonng how do you think it got added?

PatrickF1:
Ah I see it now 
Screen Shot 2021-03-23 at 3 28 08 PM

kidonng:
😅 Mea culpa, this is accidentally introduced in [`78c43a7` (#133)](https://github.com/PatrickF1/fzf.fish/pull/133/commits/78c43a75d6cc3c798e840c583c74315bc5032538), where I just copy and pasted the comment from an old commit on the web interface. I have no idea why they are added.

image

The problem here is there are actually two of them, and I only removed the latter one that broke the tests.

PatrickF1:
Is that screenshot from the web interface? I've never seen those 2 dots before. I'm guessing they are used by the GitHub JS to start and end highlighting.

kidonng:
Not, it's that problematic line pasted in Chrome Devtools, which shows invisible characters.

PatrickF1:
Thanks that makes sense! I would still love to learn what errors it was causing tho...was it fish throwing up? I looked at the CI job that broke and the main thing that broke was formatting.

kidonng:
Maybe use a older version of Fish or ask the devs, as I don't encounter any issue.

fezboy:
> I would still love to learn what errors it was causing tho...was it fish throwing up? I looked at the CI job that broke and the main thing that broke was formatting.

Here's the error: 
```
⫸  fish: Unknown command: which may include tilde, variables, etc.
in command substitution
	called on line 9 of file ~/.config/fish/functions/__fzf_search_current_dir.fish
in function '__fzf_search_current_dir'
~/.config/fish/functions/__fzf_search_current_dir.fish: Unknown error while evaluating command substitution
in function '__fzf_search_current_dir'
```
It comes up when I press "ctrl+f" I'm not sure what command that is, it was just the first thing that the README suggested doing after installing.

PatrickF1:
Thanks! Gonna release a hotfix for this.

PR #146, opened 2021-03-27
Require fish 3.2.0, update variable preview extraction

PatrickF1:
Now that fish 3.2.0 has been out for almost a month, it's time to make it the minimum required version. Thankfully, the only new expectation made by the plugin is around the new `set --show` format. It has been simplified (see https://github.com/fish-shell/fish-shell/issues/6944) such that the search shell variable preview logic, which uses `set --show`, can be greatly simplified. While this is technically backwards incompatible with fish 3.1.2, the only negative consequence for users who haven't upgraded should be that their variable preview won't be as clean and condensed but will still have all the essentials.


functions/__fzf_extract_var_info.fish
```diff
@@ -1,28 +1,18 @@
 # helper function for __fzf_search_shell_variables
 function __fzf_extract_var_info --argument-names variable_name set_show_output --description "Extract and reformat lines pertaining to \$variable_name from \$set_show_output."
-    # Extract only the lines that begin with...
-    # $variable_name: set
-    # ...or...
-    # $variable_name[
-    string match --entire --regex "^\\\$$variable_name(?:: set|\[)" <$set_show_output |
+    # Extract only the lines about the variable, all of which begin with
+    # $variable_name: or $variable_name[
+    string match --regex "^\\\$$variable_name(?::|\[).*" <$set_show_output |
```
PatrickF1, line 5: Now that the output doesn't contain lines saying "var not set in global scope", we don't have to specify "set" in the prefix to filter them out.

functions/__fzf_extract_var_info.fish
```diff
@@ -1,28 +1,18 @@
...
-
-    # Final output example for $PATH:
-    # set in global scope, unexported, with 5 elements
-    # [1] /Users/patrickf/.config/fish/functions
-    # [2] /usr/local/Cellar/fish/3.1.2/etc/fish/functions
-    # [3] /usr/local/Cellar/fish/3.1.2/share/fish/vendor_functions.d
-    # [4] /usr/local/share/fish/vendor_functions.d
-    # [5] /usr/local/Cellar/fish/3.1.2/share/fish/functions
+        #   [1] value
+        string replace --regex "^\\\$$variable_name(\[\d+\]): \|(.+)\|" '\$1 \$2'
```
PatrickF1, line 17: Now that the output doesn't contain the length string, we can simplify the middle of the two capturing groups. Furthermore, I noticed that the end of string anchor is redundant because the `.*` inside the second capturing group is greedy so I removed it.

PR #147, opened 2021-03-30
[Search files] include hidden files if current token has a leading dot

kidonng:
While it makes sense to exclude hidden files by default in #126, it actually makes the function less accessible.
Hidden files are actually being searched for quite frequently, but keeping `--hidden` in `fzf_fd_opts`  makes the changes meaningless. Something like `bind \e\cf "fzf_fd_opts=--hidden __fzf_search_current_dir"` isn't intuitive as well.

What if we include hidden files if current token has a leading dot?

- This is the same logic behind *if the current token is a directory with a trailing slash, then search will be scoped to that directory*
- This can been seen in other places, for example Alfred, which only shows hidden files if you first type a dot.

I will add test for this if you like the idea.

P.S. I know you are trying to avoid introducing more complexity to search files, and honestly I don't feel like it too. That's why this PR is **not adding new features** (like prepending `./`) but **providing sensible search defaults**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Use `fzf.fish` to interactively find and insert into the command line:
 - **Remarks**
   - prepends `./` to the selection if only one selection is made and it becomes the only token on the command line, making it easy to execute if an executable, or cd into if a directory (see [cd docs][])
   - if the current token is a directory with a trailing slash (e.g. `functions/<CURSOR>`), then search will be scoped to that directory
+  - if the current token has a leading dot (e.g. `.<CURSOR>`), then search will include hidden files
   - ignores files that are also ignored by git
   - <kbd>Tab</kbd> to multi-select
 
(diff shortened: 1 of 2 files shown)


PatrickF1:
Ok I'm sorry for being a downer. I'm going to close this for now but you are free to reopen or keep discussing with me to change my mind. In case you haven't noticed, I don't like a lot of things and eventually I change my mind after some persuasion.

kidonng:
Coming back to this PR later and I realized searching for hidden files is rarely needed (usually you already know its name), plus now I have a keybinding for searching hidden files using `fzf_fd_opts`.

PatrickF1:
Great to hear fzf_fd_opts works for this use case!

kidonng:
🤷 It's funny to see how my past self feel against this usage

PatrickF1:
Our workflows are always changing :)

PR #152, opened 2021-04-15
[Search files] fix preview for files with spaces (again)

PatrickF1:
Apparently, I didn't actually fix this bug when I tried to fix it in https://github.com/PatrickF1/fzf.fish/pull/137


PatrickF1:
@kidonng Ok this fixes bug 1 from #137 but it introduces a new bug. To repro
1. mkdir -p "normal/with spaces"
2. touch "normal/with spaces"/{1,2,3}
3. execute search directory with cursor on normal/
4. Notice you can't preview the files in normal

I realized both of these bugs actually weren't a problem until #137, so I must have broke something trying to fix something else. So confused...this is what I was afraid of when introducing all of these features to search directory.

kidonng:
(Sigh) We could have settled on [this](https://github.com/PatrickF1/fzf.fish/pull/133/commits/5fa321b295d16c32c0a87d516b5e2361bdd5c4f2) from the beginning. Thanks for the fix!

PatrickF1:
Yeah, that was my bad. I should have investigated closer. I've been pretty demotivated about maintaining this feature. It's been quite a thorn in my side siphoning my time for other feature I'd rather work on.

kidonng:
I don't want to see you feeling bad about all these issues, but I have to say this issue (Bug 2 in https://github.com/PatrickF1/fzf.fish/pull/137#issuecomment-839540256) is still no fixed.

PR #155, opened 2021-05-05
[Search files] Preview binary files using hexyl

neersighted:
This uses `hexyl`, which is a very nice (and fast!) hex viewer.

BSD `file` is also used. While the options specified are not POSIX, the almost
universally used BSD (*BSD, macOS, Linux) version (`man file.1`) accepts
these options/has compatible output. I have not tested Solaris/AIX/any other
more exotic Unix, but I suspect they are not major concerns.

Given we depend on sharkdp's other tools (fd, bat), this seems to be an easy
choice. Guards could be added to let the user know they're missing necessary
deps (or just to stub out features if the relevant tool is not installed), but
given use of fd/bat is not guarded I think it's external to this PR, and should
be added separately if deemed desirable.

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -5,7 +5,12 @@ function __fzf_preview_file --description "Print a preview for the given file ba
     set file_path $argv
 
     if test -f "$file_path" # regular file
-        bat --style=numbers --color=always "$file_path"
+        if file --mime-encoding "$file_path" | string match -rq ': binary$'
+          file --uncompress "$file_path" | fold -s
+          hexyl --border=none --color=always --length=512 "$file_path"
+        else
+          bat --style=numbers --color=always "$file_path"
+        end
     else if test -d "$file_path" # directory
         if set --query fzf_preview_dir_cmd
             # need to escape quotes to make sure eval receives file_path as a single arg


PatrickF1:
Hi @neersighted, really appreciate the idea and you even going as far to start implementing it.
Let me be honest with you though--I'm on the fence about merging this. The main reason is because I don't see what the benefit of being able to view binary is. What are your use cases for integrating with hexyl?
I do agree that if we do want to view binary, hexly, being from sharkdp, is a top choice. But as it is, adding a new feature that I suspect most users won't need and then making that one more thing for them to install, or configure, and for me to document and maintain doesn't seem worth it.
Looking forward to your reply.

kidonng:
I agree with @PatrickF1, previewing binary *may* be useful, but I don't see the point of preview *everything*.

Some thoughts, may be helpful in either this PR or anyone's fork.

- If `hexyl` is not available, failback to `xxd`, which should be widely available.
- Maybe the function can be limited to, say, gzip/brotil compressed *text files*
- Maybe try adding *image* preview, here is something I quickly made:

  <details>

  ```diff
  diff --git a/functions/__fzf_preview_file.fish b/functions/__fzf_preview_file.fish
  index 3b82157..a7b10d9 100644
  --- a/functions/__fzf_preview_file.fish
  +++ b/functions/__fzf_preview_file.fish
  @@ -5,7 +5,12 @@ function __fzf_preview_file --description "Print a preview for the given file ba
       set file_path $argv
 
       if test -f "$file_path" # regular file
  -        bat --style=numbers --color=always "$file_path"
  +        if string match -rq '\.(jpe?g|png)$' $file_path
  +            string repeat -n (math -s0 $LINES / 5) \n
  +            viu -w (math $COLUMNS - 2) -b $file_path
  +        else
  +            bat --style=numbers --color=always "$file_path"
  +        end
       else if test -d "$file_path" # directory
           if set --query fzf_preview_dir_cmd
               # need to escape quotes to make sure eval receives file_path as a single arg
  ```

  
  image

  </details>

neersighted:
The utility is mainly for me, I admit -- I've found this useful with working with domain-specific binary formats `file` does not recognize. Merely printing the mime type would be quite helpful, but a hexdump of the start of the file helps with spotting magic numbers and the like. I do realize the use here is a bit niche, but I do think a richer set of previews (like the one demoed above) add value.

PatrickF1:
Gotcha. Just curious, what are you doing that requires quickly inspecting MAGIC numbers?

Anyway, I'm sorry this is going to be very disappointing after all the hard work and enthusiasm you put into this PR, but I don't think I'm going to merge this. I don't want to call it bloat but I do think it doesn't fit my vision of the plugin as a minimalist, easy-to-understand-and-use, set of commonly-used core features.

neersighted:
Old game data with semi-documented (and decidedly non-standard) pack/archive formats, lately. Before it was embedded systems.

It was no major work at all, I already wrote this code when I used [jetrhokuan/fzf](https://github.com/jethrokuan/fzf)...

Here's a nice compromise: why not make the file preview command configurable (like the directory preview command is)? That would allow me (or anyone else) to implement whatever custom logic I want (which is what I did with the old extension), with a good sane default.

PatrickF1:
> Here's a nice compromise: why not make the file preview command configurable (like the directory preview command is)? That would allow me (or anyone else) to implement whatever custom logic I want (which is what I did with the old extension), with a good sane default.

Great idea! I've added it to my todo list. However, no promises when it'll be done. It's currently 3rd on my list, but my todo list hasn't budged in 5 months now >_<

PatrickF1:
Btw, technically, you can configure it right now by overriding `__fzf_preview_file.fish` through an auto-loaded function. Hope that'll keep you using the plugin until I implement a full solution. :)

neersighted:
Check out #156 😊

PR #156, opened 2021-05-07
[Search directory] allow customizing the file preview command

neersighted:
Based on the current implementation for directories. Seems straightforward
enough to me, but please let me know if I overlooked something.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ They are always appended last to fzf's argument list. Because fzf uses the optio
 
 ### Change the command used to preview files
 
-The search directory feature, by default, uses `bat` to preview the contents of a file. `bat` is a well-adopted `cat` replacement with syntax highlighting, line numbers, and more. If you would like to change the preview tool (to `cat` to avoid the installing a new dependency, or to add custom logic like binary or image preview), you may set the `fzf_preview_file_cmd` variable. For example, in your `config.fish`, you may put:
+The search directory feature, by default, uses `bat` to preview the contents of a file. `bat` is a well-adopted `cat` replacement with syntax highlighting, line numbers, and more. If you would like to change the preview tool (to `cat` to avoid installing a new dependency, or to add custom logic such as binary or image preview), you may set the `fzf_preview_file_cmd` variable. For example, in your `config.fish`, you may put:
 
 ```fish
 set fzf_preview_file_cmd cat
@@ -142,7 +142,7 @@ Do not specify a target path in the command, as `fzf.fish` will [prepend the fil
 
 ### Change the command used to preview folders
 
-The search directory feature, by default, uses `ls` to preview the contents of a directory. To integrate with the variety of `ls` replacements available, the command used to preview directories is configurable through the `fzf_preview_dir_cmd` variable. Set `fzf_preview_dir_cmd` in your `config.fish` similar to the above.
+The search directory feature, by default, uses `ls` to preview the contents of a directory. To integrate with the variety of `ls` replacements available (e.g. exa, lsd, tree), the command used to preview directories is configurable through the `fzf_preview_dir_cmd` variable. Set `fzf_preview_dir_cmd` in your `config.fish`:
 
 ```fish
 set fzf_preview_dir_cmd exa --all --color=always
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ They are always appended last to fzf's argument list. Because fzf uses the optio
 
 ### Change the command used to preview files
 
-The search directory feature, by default, uses `bat` to preview the contents of a file. `bat` is a well-adopted `cat` replacement with syntax highlighting, line numbers, and more. If you would like to change the preview tool (to `cat` to avoid installing a new dependency, or to add custom logic such as binary or image preview), you may set the `fzf_preview_file_cmd` variable. For example, in your `config.fish`, you may put:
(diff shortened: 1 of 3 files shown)


PatrickF1:
🤯 
No one has ever submitted a PR and actually updated the readme. Also you were so fast. THANK YOU!

README.md
```diff
@@ -130,15 +130,25 @@ They are always appended last to fzf's argument list. Because fzf uses the optio
...
 
+### Change the command used to preview files
+
+The search directory feature, by default, uses `bat` to preview the contents of a file. `bat` is a well-adopted `cat` replacement with syntax highlighting, line numbers, and more. If you would like to change the preview tool (to `cat` for availability reasons, or to add custom logic like binary or image preview), you may set the `fzf_preview_file_cmd` variable. For example, in your `config.fish`, you may put:
+
+```fish
+set fzf_preview_file_cmd cat
+```
+
+Do not specify a target path in the command, as `fzf.fish` will [prepend the directory][custom preview command] to preview to the command itself.
```
PatrickF1, line 141: s/prepend the directory/prepend the file

tests/preview_file/custom_file_preview.fish
```diff
@@ -0,0 +1,6 @@
+set multi_word_dir "tests/_resources/multi word dir"
+
+set fzf_preview_file_cmd cat
+
+set actual (__fzf_preview_file "$multi_word_dir/file 1.txt")
+@test "correctly uses custom command to preview files" "$actual" = "This is file 1"
```
PatrickF1, line 6: Hmm this is very good, but the default command, bat, does the same thing as cat...so this test wouldn't fail even if it didn't switch over to cat.

Trying to think of ideas...but I'm tired, I'll come up with something by tomorrow afternoon if you don't by then.

README.md
neersighted: Whoops!

tests/preview_file/custom_file_preview.fish
neersighted: This test actually fails with the default, as `bat` is called with `--color=always` among other options. Now I admit that is a bit sideways, but it does in fact work.

PatrickF1:
Btw, totally up to you, but you can commit normally and push. Don't need to amend and force push unless you like to :)
I don't care either way but some people like to see their commit history in their PR. And occasionally it comes in useful for big PRs when we decide on a previously passed-over solution.

neersighted:
I'm generally of the 'polished' patch set school, so I tend to restructure PRs to maintain the most logical dev history + clean application of patches.

tests/preview_file/custom_file_preview.fish
PatrickF1: Maybe something that will print the line in reverse? I thought of tac but it doesn't work b/c it reverses the order of lines, rather than chars with in a line.

PatrickF1:
Ah even better!

tests/preview_file/custom_file_preview.fish
neersighted: You could cheat with `echo`ing the path or use `sed` or something to reverse the line, or just add multiple lines and use `tac`. I think I'll leave that up to your preference.

PatrickF1: Oh, there's `rev`. It might not be available everywhere but if it's available on at least Mac and Ubuntu, then that's good enough for CI, so good enough for me.

PatrickF1: I didn't see your original comment. Your echo path idea is cogent. Go for it. We're super close to merging.

neersighted: It's in BSD as well as util-linux, so sounds good to me.

README.md
```diff
@@ -130,15 +130,25 @@ They are always appended last to fzf's argument list. Because fzf uses the optio
 - [re-populate fzf's input list on demand](https://github.com/junegunn/fzf/issues/1750)
 - change the search mode
 
+### Change the command used to preview files
+
+The search directory feature, by default, uses `bat` to preview the contents of a file. `bat` is a well-adopted `cat` replacement with syntax highlighting, line numbers, and more. If you would like to change the preview tool (to `cat` for availability reasons, or to add custom logic like binary or image preview), you may set the `fzf_preview_file_cmd` variable. For example, in your `config.fish`, you may put:
```
PatrickF1, line 135: Oh sorry small thing, can you change "to `cat` for availability reasons" to "to `cat` to avoid installing a new dependency"?

PatrickF1 (approved):

PatrickF1:
Great work! I'll clean up a little from here and merge (sorry, I'm quite obsessive with wording). Thank you SO MUCH Bjorn. Collaborating with you in a tight feedback loop has been my pleasure.

PatrickF1:
Btw, now that you have contributed such a major patch, it would mean a lot to me if you could star the repo. Thanks for considering it!

kidonng:
A little bit late, but I'd like to thank both you! This is a really great feature, I can create ［really powerful preview](https://github.com/kidonng/dotfiles/blob/master/.config/fish/functions/__fzf_preview_file_content.fish) now.

PatrickF1:
Glad you like the feature. And thank you for linking to your "really powerful preview". I'm going to add it to the cookbook WIki page. Let me know if you have reservations about that and I'll remove it.

PR #157, opened 2021-05-10
Update readme

PatrickF1:
Advertise Wiki better and as its own major entity
Point out shell variable command variable `fzf_search_vars_cmd`
Rename Wiki page FAQ to cookbook


kidonng:
The `fzf` function in https://github.com/PatrickF1/fzf.fish/wiki/Cookbook#how-can-i-open-fzf-in-a-new-tmux-pane is missing a `end`

PatrickF1:
thanks, fixed

PR #158, opened 2021-05-10
Overhaul key bindings mechanism

PatrickF1:


PatrickF1:
Hi @kidonng, could you provide your perspective and expansive CLI wisdom please? I'm wondering if the idea I have in this PR is too complicated.
Currently, there's a new function called `fzf_configure_keymap` that takes in a keymap, which is a set of pre-defined bindings, and more custom overrides, so the user can use it like `fzf_configure_keymap simple_mnemonic --directory=\e\cf`. However, the code, the tests, and the documentation is getting really complicated and I'm wondering if I should simplify it with a simple function that just binds whatever you pass it. So something like `fzf_configure_bindings --directory=\cf --variables=\cv git_log=\cl`. LMK if that makes sense.

PR #161, opened 2021-05-12
[Search directory] fix preview when base dir contains spaces (patch 3)

kidonng:
There is still, yet another bug, afflicting the search directory feature. To reproduce, `set -x fzf_preview_dir_cmd ls` and search with a base directory with spaces in its name:

image
image

As it turned out, escaping quotes isn't a good idea either. The correct way to use `eval` is to quote the code entirely. Single quotes for parameter is used for programs instead.

This is a followup from #137 and #152

### DIFF
--- a/functions/__fzf_preview_file.fish
+++ b/functions/__fzf_preview_file.fish
@@ -16,14 +16,14 @@ function __fzf_preview_file --description "Print a preview for the given file ba
     else if test -f "$file_path" # regular file
         if set --query fzf_preview_file_cmd
             # need to escape quotes to make sure eval receives file_path as a single arg
-            eval $fzf_preview_file_cmd \"$file_path\"
+            eval "$fzf_preview_file_cmd '$file_path'"
         else
             bat --style=numbers --color=always "$file_path"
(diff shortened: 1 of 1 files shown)


PatrickF1:
Can you explain why this? 
> As it turned out, escaping quotes isn't a good idea either. The correct way to use eval is to quote the code entirely.

PatrickF1:
@kidonng ^ bump on my question

kidonng:
Honestly, I don't know either. Maybe I will search for related docs or experiment a bit and give a better answer later, but I think you can merge it if you like it.

kidonng:
> Can you explain why this? 
> > As it turned out, escaping quotes isn't a good idea either. The correct way to use eval is to quote the code entirely.

I think the answer is the same as before: we need to quote the path and we need to _quote it right_. Simply scaping the quotes may looks right but in fact isn't.

PatrickF1:
> Simply escaping the quotes may looks right but in fact isn't, the eval still sees it unquoted.

Ah that makes sense.

PR #153, opened 2021-04-16
Make C-R acts as expected(redo) in vi-mode

owarai:
The latest worktree of fish-shell has reset `u` and `C-r` shortcuts in vi-mode to acts same with vim(undo, redo separately), 
fzf.fish's keybinding on `C-r` will override this redo behavior in fish vi-mode, so I change it slightly and tested on the updated version, seems works as expected,  please give a look on it.

If this modification does not solve the problem properly, feel free to close it.

related issue: https://github.com/fish-shell/fish-shell/pull/7908

PR #162, opened 2021-05-13
Only bind keys in normal mode if using `fish_default_key_bindings`

kidonng:
In Vi mode, <kbd>Ctrl</kbd> + <kbd>r</kbd> is [used for redo](https://github.com/fish-shell/fish-shell/blob/678fa2e6a9a9319f270f71ae5d3a7e14fd54bc21/share/functions/fish_vi_key_bindings.fish#L105) by default, which conflicts with `__fzf_search_history`. I believe the correct way to handle this is to only bind keys in insert mode.

Also added handler for mode switching like [autopair.fish](https://github.com/jorgebucaran/autopair.fish/blob/1222311994a0730e53d8e922a759eeda815fcb62/conf.d/autopair.fish#L7-L31).

### DIFF
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -8,21 +8,20 @@ set --global fzf_search_vars_cmd '__fzf_search_shell_variables (set --show | psu
 if not set --query fzf_fish_custom_keybindings
     function __fzf_fish_key_bindings --on-variable fish_key_bindings
         if test "$fish_key_bindings" = fish_default_key_bindings
-            # \cf is Ctrl+f
-            bind \cf __fzf_search_current_dir
-            bind \cr __fzf_search_history
-            bind \cv $fzf_search_vars_cmd
-            # The following two key binding use Alt as an additional modifier key to avoid conflicts
-            bind \e\cl __fzf_search_git_log
-            bind \e\cs __fzf_search_git_status
+            set modes default insert
         else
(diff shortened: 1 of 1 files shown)


PatrickF1:
Interesting, there's more demand for this than I thought. This was initially proposed in #153. Can you take a look at the work I started in #158 to see if it'll be adequate to address this issue?

kidonng:
I saw that PR but forgot it when I made this. Will take a look later

PatrickF1:
@kidonng how come you closed this PR?

kidonng:
I haven't got the time to see #158, but since this is an issue already being worked on, I think it's better to close it. If I see it necessary we can reopen later.

PatrickF1:
I realized #158 won't fix this, though it'll make it easier for users to workaround it. Yeah let's just explore this when Ctrl+R = redo on the next release of fish. I don't want to make updates for fish master branch.

kidonng:
> I don't want to make updates for fish master branch.

That's a point I forgot to mention. Though it doesn't hold water, one don't have to actually use `master` branch of fish, they just need to drop the latest `fish_vi_key_bindings.fish` into `~/.config/fish/functions`. In fact this is what I do personally.

(It's a bless to be able to easily override/backport latest feature/fix into fish, thanks to the autoload feature.)

PR #163, opened 2021-05-20
[Search directory] fix specifying base directories using quotes

kidonng:
This was presented in the [initial PR](https://github.com/PatrickF1/fzf.fish/pull/75) when the feature is added, but was (possibly mistakenly) broken by #137.

### DIFF
--- a/functions/__fzf_search_current_dir.fish
+++ b/functions/__fzf_search_current_dir.fish
@@ -13,7 +13,7 @@ function __fzf_search_current_dir --description "Search the current directory. R
 
     # If the current token is a directory and has a trailing slash,
     # then use it as fd's base directory.
-    if string match --quiet -- "*/" $token && test -d "$unescaped_exp_token"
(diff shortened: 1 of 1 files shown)


PatrickF1:
How can I repro the bug?

kidonng:
I don't consider it a bug, though technically it _did_ stop working. 

To reproduce, just input `'with space/` and trigger file search. It won't scope the search to the folder.

PatrickF1:
Hmm I was able to repro. But you know what? I vaguely remember changing `unescaped_exp_token` to just `token` b/c it broke something else. Now trying to remember what that was...

kidonng:
> I vaguely remember changing `unescaped_exp_token` to just `token` b/c it broke something else.

If it breaks anything we should fix it instead. However this only affects the string match and I don't see any other issue.

PatrickF1:
Did some more testing and it seems to not break anything. If it does introduce a new bug, I trust you to find it 😉

PR #164, opened 2021-05-20
[Search directory] fix previewing symlinks containing spaces

kidonng:
Due to #152, this is broken while all other `$file_path` references have been correctly quoted.

image


PatrickF1 (approved):
Thanks for the fix. Sorry for the delay.

PR #165, opened 2021-05-27
[Search variables] enable selecting multiple variables

kidonng:
This feature started with the discussion in https://github.com/PatrickF1/fzf.fish/pull/73#discussion_r552143666.

After such a long time I'm convinced that keeping `--multi` in `FZF_DEFAULT_OPTS` is not a good idea, but we can make more functions allow multiple selections by default.

IMO, out of 5 current fzf.fish functions, only `__fzf_search_history` does not need `--multi`. `__fzf_search_current_dir` and `__fzf_search_git_status` already has `--multi` by default, which is natural since they deal with files.

`__fzf_search_shell_variables` joins the party in this PR, the most prominent usage I can think of now is `set -e` multiple variables.

As for `__fzf_search_git_log`, that will be in a separate PR since we may need to discuss how to handle multiple selections there (it's not as simple as `string join " "`).

### DIFF
--- a/functions/__fzf_search_shell_variables.fish
+++ b/functions/__fzf_search_shell_variables.fish
@@ -31,7 +31,8 @@ function __fzf_search_shell_variables --argument-names set_show_output set_names
 
     set variable_names_selected (
         printf '%s\n' $all_variable_names |
-        fzf --multi --preview "__fzf_extract_var_info {} $set_show_output" \
+        fzf --preview "__fzf_extract_var_info {} $set_show_output" \
+            --multi \
             --query=$cleaned_curr_token \
             $fzf_shell_vars_opts
     )
--- a/functions/__fzf_search_shell_variables.fish
+++ b/functions/__fzf_search_shell_variables.fish
@@ -39,9 +39,10 @@ function __fzf_search_shell_variables --argument-names set_show_output set_names
     if test $status -eq 0
         # If the current token begins with a $, do not overwrite the $ when
         # replacing the current token with the selected variable.
+        # Uses brace expansion to prepend $ to each variable name.
         commandline --current-token --replace (
             if string match --quiet -- '$*' $current_token
-                string join " " \$$variable_names_selected
+                string join " " \${$variable_names_selected}
             else
                 string join " " $variable_names_selected
             end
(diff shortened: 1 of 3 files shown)


functions/__fzf_search_shell_variables.fish
```diff
@@ -29,21 +29,23 @@ function __fzf_search_shell_variables --argument-names set_show_output set_names
...
 
     if test $status -eq 0
         # If the current token begins with a $, do not overwrite the $ when
         # replacing the current token with the selected variable.
-        if string match --quiet -- '$*' $current_token
-            commandline --current-token --replace \$$variable_name
-        else
-            commandline --current-token --replace $variable_name
-        end
+        commandline --current-token --replace (
```
kidonng, line 42: We can set a variable in the `if` and `commandline --current-token --replace $variable` instead, but I found this method more readable, and personally speaking, more beautiful.

PatrickF1: can you clarify what the other option you had in mind is?

PatrickF1:
Sorry been busy working on the other PR. I like this idea and will eventually merge this but I don't like the idea of selecting multiple variables with $ and having the commandline become $var1 $var2. Seems like this would never be used and also users wouldn't really expect it. I propose adding this feature but removing the $ feature. What do you think?

functions/__fzf_search_shell_variables.fish
PatrickF1: oh I get it, you mean set the variable to be written to the commandline in the if block. yeah I like this way better as well

PatrickF1:
>  Seems like this would never be used and also users wouldn't really expect it. I propose adding this feature but removing the $ feature.

Never mind, I can see this being used not-so-rarely to conveniently select multiple related variables

kidonng:
> but I don't like the idea of selecting multiple variables with $ and having the commandline become $var1 $var2. Seems like this would never be used and also users wouldn't really expect it.

Yeah this is a side effect of this PR. We actually have three options:

- Don't handle `$` at all, leave it be
- Only handle `$` if made one selection
- Handle `$` in all situations

The first one is also the simplest, and the second one is the *smartest*. But I will go for the third one:
- I have contributed many features to fzf.fish. I really like the idea that fzf.fish should not do too much and be a generic tool. But being generic doesn't mean you have to be **dump**, you can still be smart enough and provide sensible defaults. That's why I brought up the `$` handler.
- I agree `$1 $2` will never be used in practice, but being too smart will also break user's expectation. And that will be an extra quirk to document as well. I think you can treat is as some sort of [parameter expansion](https://fishshell.com/docs/current/index.html#expand):

  ```fish
  set selection 1 2
  echo '$'$selection // $1 $2
  ```

PR #175, opened 2021-06-15
Fix default key sequence doc in fzf_configure_bindings

jonashaag:
I got confused by this; I am assuming `fzf_configure_bindings --help` should show the default key bindings, which were different from what was documented.


PatrickF1 (approved):
Oh my! 🤭 I did mess that up didn't it? Good catch and thank you for submitting this!

PR #177, opened 2021-06-16
Define command completions for fzf_configure_bindings

PatrickF1:


PatrickF1:
todo: consider whether to add a set of completions for --option=

PatrickF1:
Decided against it:
- to keep the tab completion menu from getting too packed
- there are almost no examples of completions that include = at the end of the option in fish's official completions
- the = completions don't work well, it still adds a space after selecting the option, suggesting to the user that they can include a space after the option even though it's required

PR #178, opened 2021-06-18
[Search directory] fix quick directory change for hidden directories

### LINKED ISSUE #171: Auto prepend ./ for quick directory change doesn't work if directory starts with .
**Describe the bug**
When selecting a directory that begins with `.`, the result is not prepended with `./` and `fish` gives an error: `fish: Unknown command: .config`.

**To Reproduce**
1) `set -x fzf_fd_opts -H`
2) `cd` to a folder with folders beginning with `.`
3) `C-f` to trigger `__fzf_search_current_dir`
4) Select a folder such as `.config` that begins with a `.` and press `enter` to select it
5) Press `enter` again

**Expected behavior**
Prepend with `./` so that the second `enter` does a `cd` into the folder

**Screenshots**
N/A

**Environment**

- Fish: `3.1.2`
- fzf.fish: not sure how to check but I did run `fisher update` prior to posting this bug so it should be latest as of this bug report
- plugin manager: `fisher 4.3.0`
- terminal: `kitty`
- OS: `debian-based`

**Additional context**
This appears to have been introduced by b19e3f82 which has a regex check for a path used as the base directory and as a side effect explicitly does not add the `./` to folders beginning with `.`.

Laptop765:
Fixes #171 by making `.` optional but `/` required when deciding not to prepend `./`.

### DIFF
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -34,7 +34,8 @@ function _fzf_search_directory --description "Search the current directory. Repl
         # Then, the user only needs to hit Enter once more to potentially cd into or execute that path.
         if test (count $file_paths_selected) = 1
             set commandline_tokens (commandline --tokenize)
-            if test "$commandline_tokens" = "$token"
+            set file_selected $file_paths_selected[1]
+            if test "$commandline_tokens" = "$token" -a -d "$file_selected"
                 set file_paths_selected ./$file_paths_selected
             end
         end
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -33,7 +33,6 @@ function _fzf_search_directory --description "Search the current directory. Repl
         # - the user was in the middle of inputting the first token,
         # Then, the user only needs to hit Enter once more to potentially cd into or execute that path.
         if test (count $file_paths_selected) = 1
-                && not string match --quiet --regex "^\.?/" $file_paths_selected
             set commandline_tokens (commandline --tokenize)
             if test "$commandline_tokens" = "$current_token"
                 set file_paths_selected ./$file_paths_selected
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -34,7 +34,7 @@ function _fzf_search_directory --description "Search the current directory. Repl
         # - and the path doesn't already begin with a dot or slash
         # Then, the user only needs to hit Enter once more to potentially cd into or execute that path.
         if test (count $file_paths_selected) = 1 \
-                && not string match --quiet --regex "^[.|/]" $file_paths_selected
+                && not string match --quiet --regex "^\.?/" $file_paths_selected
             set commandline_tokens (commandline --tokenize)
             if test "$commandline_tokens" = "$current_token"
                 set file_paths_selected ./$file_paths_selected
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -31,7 +31,6 @@ function _fzf_search_directory --description "Search the current directory. Repl
         # automatically prepend ./ to the selected path if
         # - only one path was selected,
         # - the user was in the middle of inputting the first token,
-        # - and the path doesn't already begin with a / or ./
         # Then, the user only needs to hit Enter once more to potentially cd into or execute that path.
         if test (count $file_paths_selected) = 1 \
                 && not string match --quiet --regex "^\.?/" $file_paths_selected
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -32,7 +32,7 @@ function _fzf_search_directory --description "Search the current directory. Repl
         # - only one path was selected,
         # - the user was in the middle of inputting the first token,
         # Then, the user only needs to hit Enter once more to potentially cd into or execute that path.
-        if test (count $file_paths_selected) = 1 \
+        if test (count $file_paths_selected) = 1
                 && not string match --quiet --regex "^\.?/" $file_paths_selected
             set commandline_tokens (commandline --tokenize)
             if test "$commandline_tokens" = "$current_token"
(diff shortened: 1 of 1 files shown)


Laptop765:
I'd have added tests but the machine I'm on is currently locked at `fish 3.1.2` so I can't actually run/test against `HEAD`.  I have a fork of `v6.5` that I made these same changes on and seems to be working as expected.

PatrickF1:
Thanks for the PR!

Laptop765:
Happy to help!

kidonng:
I think some later changes (especially #133) makes #119 unnecessary. In other words, we no longer need to check the selected paths again because they never start with literal tilde (`~`) and such.

functions/_fzf_search_directory.fish
```diff
@@ -31,10 +31,10 @@ function _fzf_search_directory --description "Search the current directory. Repl
         # automatically prepend ./ to the selected path if
         # - only one path was selected,
         # - the user was in the middle of inputting the first token,
-        # - and the path doesn't already begin with a dot or slash
+        # - and the path doesn't already begin with a / or ./
```
kidonng, line 34:
```suggestion
```

functions/_fzf_search_directory.fish
```diff
@@ -31,10 +31,10 @@ function _fzf_search_directory --description "Search the current directory. Repl
...
         # Then, the user only needs to hit Enter once more to potentially cd into or execute that path.
         if test (count $file_paths_selected) = 1 \
```
kidonng, line 36:
```suggestion
        if test (count $file_paths_selected) = 1 \
        if test (count $file_paths_selected) = 1
```

functions/_fzf_search_directory.fish
```diff
@@ -31,10 +31,10 @@ function _fzf_search_directory --description "Search the current directory. Repl
...
-                && not string match --quiet --regex "^[.|/]" $file_paths_selected
+                && not string match --quiet --regex "^\.?/" $file_paths_selected
```
kidonng, line 37:
```suggestion
```

PatrickF1:
Sorry for the delay. This is on my radar but I've been busy or sick. I'll get this merged eventually

Laptop765:
No worries, take care of yourself!  I have a fork (and am stuck on an older `fish` at work so I'm on `v6.5` anyway) so I'm not blocked; I just wanted to contribute back.

PatrickF1:
Something is off. The plugin doesn't prepend ./ anymore when there is a commandline token. E.g. if you start typing in Doc and you select Documents, you'd expect ./Documents but it outputs Documents.

Laptop765:
Ahh good catch!

It looks like when nothing is typed, both `$commandline_tokens` and `$current_token` are blank whereas if something is typed before activating the search `$commandline_tokens` contains what was typed but `$current_token` is still blank.  I'm not sure where `$current_token` is coming from.

I wonder if it might be better to do something like `test -d $file_paths_selected` to check if the selection is a directory rather than the current approach?

Laptop765:
I think it might be missing `set current_token (commandline --current-token)` after the `set commandline_tokens`?  I'm pretty new to `fish` so I'm not fully sure what this code is supposed to be doing or if that's a correct fix.

Laptop765:
If there are two tokens, what is the expectation of completing a directory?  Activation happens where the `|` is in the below examples representing the cursor position
  * First token: `Doc| test`
  * Subsequent token: `test Doc|`

Is the expectation that it should complete as `./` or no?  If yes then I think we can remove the `test "$commandline_tokens" = "$current_token"` check.  If no, we can probably do `test (count $commandline_tokens)` instead and remove `$current_token`.

kidonng:
There used to be a `set current_token (commandline --current-token)`, which is removed in [`160e862` (#119)](https://github.com/PatrickF1/fzf.fish/pull/119/commits/160e8620c0dfeea7b84dc7d3b836c6fc248b6e2a).

PatrickF1:
> I wonder if it might be better to do something like test -d $file_paths_selected to check if the selection is a directory rather than the current approach?

Yeah, this was the old approach. I forgot why we changed it. I'd be open to going back and just append a / to all selected directories. It be a little un-performant if selecting tons of results, but that should be rare @kidonng any thoughts? 

> Is the expectation that it should complete as ./Documents in both cases or no?

I only wanted to prepend ./ to the directory for the purposes of when the user was looking for exactly one folder to cd to (if you hit enter when the only token is ./directory, it changes to directory). So no. And we can't just do `test (count $commandline_tokens) = 1` because if the user has `something |`, and they select documents, they wouldn't be able to hit enter to cd into whatever directory was selected.

> There used to be a set current_token (commandline --current-token), and was removed in 160e862 (#119), for no good reason.

Good catch! Thanks!

PatrickF1:
> In other words, we no longer need to check the selected paths again because they never start with literal tilde (~) and such.

Actually, ~ can expand to `/Users/me`, in which case we still need the check.

kidonng:
I posted that comment because I did some tests by hand. But perhaps I should give it another shot and post the results here.

kidonng:
[My previous comment](https://github.com/PatrickF1/fzf.fish/pull/178#issuecomment-865433958) is incorrect because the `./` prepending never happens due to the missing variable.

@Laptop765 sorry for the misguidance. You need to revert my suggestions.

kidonng:
> > I wonder if it might be better to do something like test -d $file_paths_selected to check if the selection is a directory rather than the current approach?
> 
> Yeah, this was the old approach. I forgot why we changed it. I'd be open to going back and just append a / to all selected directories. It be a little un-performant if selecting tons of results, but that should be rare @kidonng any thoughts? 

"The old approach" never entered the codebase and stayed at the original PR. No need to revert since we have no issues with current _approach_ (code is a little broken though).

PatrickF1:
Ok sorry for the delay. I figured out a good way to fix this. Unfortunately, it's gonna be a decent amount of work so probably better if I start a new PR.

PR #180, opened 2021-06-21
Skip config if not in interactive mode

kidonng:
All of fzf.fish's features are only relevant and helpful in interactive mode. However, its config gets sourced even in non-interactive mode. So, let's skip configuration in non-interactive mode to make non-interactive fish shells a bit faster. This greatly benefits other plugins as well; for example [tide](https://github.com/IlanCosman/tide) makes use of non-interactive fish shells extensively.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -28,7 +28,7 @@ jobs:
 
       - name: Run full test suite
         run: fishtape tests/*/*.fish
-        shell: fish --interactive {0}
+        shell: fish {0}
         # timeout in case tests get stuck on fzf
         timeout-minutes: 3
 
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -28,7 +28,7 @@ jobs:
 
       - name: Run full test suite
         run: fishtape tests/*/*.fish
-        shell: fish {0}
+        shell: fish --interactive {0}
         # timeout in case tests get stuck on fzf
         timeout-minutes: 3
 
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -5,7 +5,12 @@
 set --global _fzf_search_vars_command '_fzf_search_variables (set --show | psub) (set --names | psub)'
 
 # Skip rest of the config if not in interactive mode to speed shell startup a little
-status is-interactive || exit
+# Check for $CI environment variable [1] before fishtape issue [2] is resolved
+# [1] https://github.com/jorgebucaran/fishtape/issues/63
+# [2] https://docs.github.com/actions/reference/environment-variables#default-environment-variables
+if not status is-interactive && test "$CI" != true
+    exit
+end
 
 # Install the default bindings, which are mnemonic and minimally conflict with fish's preset bindings
 fzf_configure_bindings
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -6,8 +6,8 @@ set --global _fzf_search_vars_command '_fzf_search_variables (set --show | psub)
 
 # Skip rest of the config if not in interactive mode to speed up shell startup
 # Check for $CI environment variable [1] before fishtape issue [2] is resolved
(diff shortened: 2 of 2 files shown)


PatrickF1:
Great idea! I will merge this as soon as I figure out how to make it still possible to run tests locally. B/c right tests only work on CI but I run them locally when I'm developing.

kidonng:
Why not just pretend you are a CI (`set -x CI true`?)

PatrickF1:
Sure that'll work for me.

PatrickF1:
Actually, the tests still fail for me locally even with CI=true and set -x CI true

PatrickF1:
@kidonng can you help me answer these questions
- I did s/subshell/session on your PR's description b/c it seems fish doesn't have subshells. Is this correct?
- where in tide's code are non-interactive sessions run?
- does this change affect command substitution?
- where else are non-interactive sessions run?

kidonng:
> * I did s/subshell/session on your PR's description b/c it seems fish doesn't have subshells. Is this correct?

[Subshell](https://tldp.org/LDP/abs/html/subshells.html) here is literally just shell as a subprocess. You can call it subprocess if you don't like it, but session is something entirely different.

> * where in tide's code are non-interactive sessions run?

[`fish_prompt`](https://github.com/IlanCosman/tide/blob/630ae9f7d93c5f53880e7d59ae4e61f6390b71a1/functions/fish_prompt.fish#L21-L26). tide's code is straightforward so I guess I don't need to repeat it here.

> * does this change affect command substitution?
> * where else are non-interactive sessions run?

I'm not sure what you mean or what are referring to, and honestly I think they are out of scope of this PR.

PatrickF1:
> I'm not sure what you mean or what are you referring to, and honestly I think they are out of scope for this PR.

I'm trying to learn these aspects of fish so I can properly write a summary of how this affects users' shell performance.

PatrickF1:
What is fish session then?

PatrickF1:
Ah funny that https://github.com/PatrickF1/fzf.fish/issues/181 was created the same day. Now I can add another reason this change is good: it makes fzf previews load a bit faster!

kidonng:
> > I'm not sure what you mean or what are you referring to, and honestly I think they are out of scope for this PR.
> 
> I'm trying to learn these aspects of fish so I can properly write a summary of how this affects users' shell performance.

🤔 Perhaps I should just give a short reply to those questions:

> - does this change affect command substitution?

That depends on what do you mean by "command substitution". But from my view, you don't usually start another fish in command substitution so no.

> - where else are non-interactive sessions run?

Again, are you referring to fzf.fish, tide or fish in general? I don't think I can give a better answer than reading the code/docs directly.

kidonng:
> What is fish session then?

From starting a fish shell to exit this particular shell. In fact that's not a fish  terminology, not even a shell terminology, it's just session's  literal meaning.
There is something that I want to note, though. I consider "session" to imply an _interactive_ launch. You don't think your shell scripts "start many sessions" right?

kidonng:
> Ah funny that https://github.com/PatrickF1/fzf.fish/issues/181 was created the same day. I realized another reason this change is good: it makes fzf previews load a bit faster!

Hmm I don't think that issue has anything to with this change. Glad you are happy about it though.

PatrickF1:
Thanks for answering my questions.

> Hmm I don't think that issue has anything to do with this change. Glad you are happy about it though.

It does doesn't it? When fzf loads a previews, it executes a new fish shell, which means it has to load the conf.d/fzf.fish in the process.

kidonng:
> It does doesn't it? When fzf loads a previews, it executes a new fish shell, which means it has to load the conf.d/fzf.fish in the process.

**That issue #181** has nothing to do with preview/shell launch speed.

The faster preview has been mentioned from the very beginning in this PR's description, until it is edited:

image

PatrickF1:
> The faster preview has been mentioned from the very beginning in this PR's description, until it is edited

😅 Oops my bad. I think I was not mentally all there when I edited it. Sorry !

PR #185, opened 2021-07-04
[Search directory] Append / instead of prepending ./ for quick cd

### LINKED ISSUE #171: Auto prepend ./ for quick directory change doesn't work if directory starts with .

PatrickF1:
I decided appending / is a better, cleaner solution for quick cd than appending ./
- since prepending ./ is causing issues when the path starts with . or /, why not just append /? fd never appends / to its output so this is safe to do (see https://github.com/sharkdp/fd/issues/436)
- of course, we can't just append / to everything; we need to test if the path is actually a directory, but since we only do this when there's one path selected, it's not performance hindering
- a / at the end of only directories is cleaner, natural, and has a more comprehensible than prepending ./ to anything even when it's not needed
- the code and documentation become much more straightforward
- this will remove support for executables, which cuts back on product debt
- fixes #171 (is compatible with hidden directories)

This is about the same as the PR that started this all (#72) except we're using a trailing / instead.

Do note that symlink behavior changes if the directory has a trailing / (https://github.com/PatrickF1/fzf.fish/pull/185#issuecomment-873533385).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Use `fzf.fish` to interactively find and insert different shell entities into th
 - **Key binding and mnemonic:** <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>F</kbd> (`F` for file)
 - **Preview window:** file with syntax highlighting, directory contents, or file type
 - **Remarks**
-  - prepends `./` to the selection if only one selection is made and it becomes the only token on the command line, making it easy to execute if an executable, or cd into if a directory (see [cd docs][])
+  - appends `/` to the selection if it is a directory and the only path selected, making it quick to cd into (see [cd docs][])
   - if the current token is a directory with a trailing slash (e.g. `.config/<CURSOR>`), then that directory is searched instead
   - ignores files that are also ignored by git
   - <kbd>Tab</kbd> to multi-select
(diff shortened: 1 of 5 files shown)


kidonng:
Also read [the original PR](https://github.com/PatrickF1/fzf.fish/pull/72#issuecomment-753743213) for why I changed my mind to prepend `./` instead of appending `/`.

PatrickF1:
> Also read the original PR for why I changed my mind to prepend ./ instead of appending /.

Thanks. I replied in #171. Let's keep the discussion there.

PR #186, opened 2021-07-05
Fix bind error messages appearing in non-interactive windows

### LINKED ISSUE #183: fzf results contain bind commands
## Describe the bug
adding `fzf_configure_bindings --directory=\cf` to `config.fish` adds additional lines of text to all fzf results:

`bind --preset \\cv fish_clipboard_paste`
`bind --preset -M insert \\cv fish_clipboard_paste`

## Screenshots
example1

example2

image

## Environment

- [x ] I have made sure my CLIs meet the [minimum version requirements](https://github.com/PatrickF1/fzf.fish#installation)

Versions installed:

- Fish: [e.g. 3.2.2] Fish Shell 3.3.0
- fzf.fish: [e.g. 7.0] check
- plugin manager: [e.g. fisher 4.3.0] check
- terminal: [e.g. iTerm] ITerm 3.48
- OS: [e.g. macOS 10.15.7] macOS BigSur 11.4

## Additional context

Anything else?

kidonng:
This fixes #183, which is a resulting bug from #180.

## Explanation of the issue

#180 makes `$_fzf_search_vars_command` unavailable for non-interactive sessions. However, `fzf_configure_bindings` depends on that variable being available. If the user puts `fzf_configure_bindings` in their `config.fish`, `fzf_configure_bindings` gets executed in non-interactive sessions e.g. fzf preview windows or by tide, and the following line in the function...

https://github.com/PatrickF1/fzf.fish/blob/17d54b576919ee77644779db2dcae56d372a8830/functions/fzf_configure_bindings.fish#L33

...essentially becomes: 

```fish
test -n $key_sequences[5] && bind --mode $mode $key_sequences[5]
```

which results in error message like `bind --preset \cv fish_clipboard_paste` or `bind: No binding found for sequence '\cv'`.

## Solution
Don't execute the entirety of `fzf_configure_bindings` if not in interactive mode. And for good measure, quote `_fzf_search_vars_command` so that even if it does get executed, the bind command will run successfully.

### DIFF
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -1,14 +1,13 @@
-# fzf.fish is only meant to be used in interactive mode. If not in interactive mode and not in CI, skip the config to speed up shell startup
-if not status is-interactive && test "$CI" != true
-    exit
-end
-
 # Because of scoping rules, to capture the shell variables exactly as they are, we must read
 # them before even executing _fzf_search_variables. We use psub to store the
 # variables' info in temporary files and pass in the filenames as arguments.
 # # This variable is global so that it can be referenced by fzf_configure_bindings and in tests
 set --global _fzf_search_vars_command '_fzf_search_variables (set --show | psub) (set --names | psub)'
 
+# fzf.fish is only meant to be used in interactive mode. If not in interactive mode and not in CI, skip the config to speed up shell startup
+if not status is-interactive && test "$CI" != true
+    exit
+end
 
 # Install the default bindings, which are mnemonic and minimally conflict with fish's preset bindings
 fzf_configure_bindings
--- a/conf.d/fzf.fish
+++ b/conf.d/fzf.fish
@@ -1,13 +1,14 @@
+# fzf.fish is only meant to be used in interactive mode. If not in interactive mode and not in CI, skip the config to speed up shell startup
+if not status is-interactive && test "$CI" != true
+    exit
+end
+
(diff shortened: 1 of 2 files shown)


PatrickF1:
Hi, thanks for discovering the root of the bug! Very clever, I was kinda stuck.
Wouldn't it be better to just add `status -i || exit` in fzf_configure_bindings?

PatrickF1:
Yeah, fish doesn't even bind any of the preset bindings in non-interactive:
```
fish -c '                                                                                                       
         echo testing non-interactive mode bindings
         bind --user
     '
testing non-interactive mode bindings
```
So I think it makes sense for it to just do nothing.

PatrickF1:
Ok I'm just going to go with adding `status -i || exit` to the `configure_fzf_bindings` rather than setting `_fzf_search_vars_command` even when in non-interactive. Reason being
1. `configure_fzf_bindings` doesn't need to run in non-interactive anyway
2. fish doesn't even bind any of the preset bindings in non-interactive
3. this is the most performance optimal: it reduces the amount of code executed in both `conf.d/fzf.fish` and `configure_fzf_bindings` when in non-interactive mode

kidonng:
Sorry for the delay! I was going to reply right after your last comment, but somehow it just got put off.

> Wouldn't it be better to just add `status -i || exit` in fzf_configure_bindings?

Sure that seems reasonable.

PR #188, opened 2021-07-12
[RFC] Add support for z

rominf:
Hi. I've created support for https://github.com/jethrokuan/z. I did it very quickly, in about an hour. Sorry, I've read the contribution guide too lately. So, please let me know if I should write tests for this, or will it be discarded anyway for some reason (then I'll move this code to a separate plugin). GIF is also missing, of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ Use `fzf.fish` to interactively find and insert different shell entities into th
 
 _The prompt used in the screencasts was created using [IlanCosman/tide][]._
 
+### Most visited directories
+
+- **Search input:** listing of most visited directories, provided by [z][]
+- **Key binding and mnemonic:** <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Z</kbd> (`Z` for z)
+- **Preview window:** directory contents
+
 ## Installation
 
 First, install a proper version of these CLIs:
@@ -71,6 +77,7 @@ First, install a proper version of these CLIs:
 | [fzf][]  | 0.27.2                   | command-line fuzzy finder that powers this plugin |
 | [fd][]   | 7.5.0                    | much faster and friendlier alternative to `find`  |
 | [bat][]  | 0.16.0                   | smarter `cat` with syntax highlighting            |
+| [z][]    | <2.7.0                   | pure-fish z directory jumping                     |
 
 On certain distribution of Linux, you need to alias `fdfind` to `fd` (see [#93](https://github.com/PatrickF1/fzf.fish/discussions/93)).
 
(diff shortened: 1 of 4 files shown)


PatrickF1:
Hey Roman, thanks for the PR! This is a very interesting and clever idea, but I'm afraid it's not a good for this plugin for these reasons
1. Not everyone uses z, there are many alternatives such as https://github.com/ajeetdsouza/zoxide, https://github.com/b4b4r07/enhancd, https://github.com/oh-my-fish/plugin-z (though I admit I've only seen 2-3 people use this one). I only want to add features that people can and will use. You might wonder why I require `fd` and `bat` then. Well, the user only has to install it, but everyone has a need for searching and previewing files quickly.
2. I actually slightly regret adding a dependency for fd and bat and I don't want to add more dependencies.

Sorry to be a wet blanket! Sorry that you already put in so much work. Good luck making this a separate plugin :)

rominf:
Hi Patrick!

Is your word final? Because I have ideas, how to satisfy almost everyone :)

First of all, it's possible to build this "search in directory history" upon this built-in fish command: https://fishshell.com/docs/current/cmds/dirh.html#cmd-dirh by default. So, no extra dependencies are needed.

And second, what is related to this, is that all these "better cd" programs share in common that they generate a list of directories. Actually, the only thing we need for this plugin is to get this list, so considering that the number of popular "better cd" programs is limited, I would just ask the user to define a variable: `fzf_cd_provider` (one of `fish-shell/fish-shell` (the default) `ajeetdsouza/zoxide`, `jethrokuan/z`, `skywind3000/z.lua`, etc.). If this variable is not defined (by default), the feature is disabled, otherwise, the plugin just selects the appropriate command to extract the list of directories and handles this list the same way for all providers. Hence, we get:
1. No extra hard dependencies, but still useful behavior by default.
2. Support for multiple "better cd" programs, which can be selected using just one intuitive variable.
3. Not so much code for supporting this functionality.
If the user uses something really rare, he/she can just overwrite a function for getting the list of directories.

I volunteer to implement everything. What do you think?

PatrickF1:
Hi Roman!
Wow I really appreciate the enthusiasm! What you described actually sounds very viable and well-engineered except the fact that the formats of the files might be different, so you'd have to parse them differently, and possibly even from binary.
However, one thing I should have mentioned actually, is I have low interest in this feature right now. Each new feature added to this plugin is more surface area for bugs, question, maintenance, new development, and I'm pretty strapped for time as it is.

Two consolations I can give you
1. I might change my mind later when my personal life gets less busy, but that might be a year from now
2. if you make your own plugin, I'd be happy to mention it somewhere in the project to give you publicity :)

PR #194, opened 2021-08-15
Set SHELL and FZF_DEFAULT_OPTS in wrapper around fzf

PatrickF1:
Create a wrapper around internal execution of fzf, and in it set `SHELL` and `FZF_DEFAULT_OPTS`. This has three huge advantages:
1. Before `FZF_DEFAULT_OPTS` was set in config.fish. By moving that code into an internal wrapper, we no longer meaningfully dirty the shell environment. This means fewer side-effects as a result of installing this plugin and so feels a lot purer.
2. This eliminates the problem of properly uninstalling the default `FZF_DEFAULT_OPTS` in the uninstall function.
2. We have finally consolidated of the code that forces `SHELL` to be fish.

### DIFF
--- a/functions/_fzf_configure_bindings_help.fish
+++ b/functions/_fzf_configure_bindings_help.fish
@@ -21,7 +21,7 @@ DESCRIPTION
     sequences. Rather, consider using fish_key_reader to manually validate them.
 
     In terms of experimentation, fzf_configure_bindings erases any bindings it previously installed
-    before installing new ones so it can be repeatedly invoked in the same fish session without
+    before installing new ones so it can be repeatedly executed in the same fish session without
     problem. Once the desired fzf_configure_bindings command has been found, add it to config.fish
     in order to persist the bindings.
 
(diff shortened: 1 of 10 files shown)


functions/_fzf_configure_bindings_help.fish
```diff
@@ -21,7 +21,7 @@ DESCRIPTION
     sequences. Rather, consider using fish_key_reader to manually validate them.
 
     In terms of experimentation, fzf_configure_bindings erases any bindings it previously installed
-    before installing new ones so it can be repeatedly invoked in the same fish session without
+    before installing new ones so it can be repeatedly executed in the same fish session without
```
PatrickF1, line 24: out of scope change: simply using consistent terminology of execute a function instead of invoke a function

functions/_fzf_search_git_log.fish
```diff
@@ -2,16 +2,12 @@ function _fzf_search_git_log --description "Search the output of git log and pre
     if not git rev-parse --git-dir >/dev/null 2>&1
         echo '_fzf_search_git_log: Not in a git repository.' >&2
     else
-        # Make sure that fzf uses fish to execute git show.
-        # See similar comment in _fzf_search_variables.fish.
-        set --local --export SHELL (command --search fish)
```
PatrickF1, line 7: this was actually unnecessary, forgot to remove it in a previous PR

functions/_fzf_search_history.fish
```diff
@@ -2,14 +2,10 @@ function _fzf_search_history --description "Search command history. Replace the
     # history merge incorporates history changes from other fish sessions
     builtin history merge
 
-    # Make sure that fzf uses fish so we can run fish_indent.
-    # See similar comment in _fzf_search_variables.fish.
-    set --local --export SHELL (command --search fish)
```
PatrickF1, line 7: this was also unnecessary b/c fish_indent is a binary that is most likely available in other shells

PR #199, opened 2021-09-05
Remove `$_fzf_search_vars_command`

kidonng:
See #198

This also removes the CI check since it's no longer needed.


PatrickF1:
As I explained in the discussion, I would prefer to keep the variable there. However, the rest of this PR regarding removing the CI check is valuable and if you remove everything but that, I'll merge this. Thanks!

PR #209, opened 2021-11-23
[Search variables] Remove excessive escape for replacement

rmgk:
This fixes displaying the value of variables in the variable preview.

This change confuses me for the following observations:
escaping the $ inside the regex should never work as intended (as far as I understand), but using fish 3.3.1 the existing code somehow works in the first fish instance I start after deleting the fish_variables file.
The changed code always works for me.

### DIFF
--- a/functions/_fzf_extract_var_info.fish
+++ b/functions/_fzf_extract_var_info.fish
@@ -11,5 +11,5 @@ function _fzf_extract_var_info --argument-names variable_name set_show_output --
         #   [1]: |value|
         # ...with...
         #   [1] value
-        string replace --regex ": \|(.*)\|" ' \$1'
+        string replace --regex ": \|(.*)\|" ' $1'
 end


PatrickF1:
Hi @rmgk thanks for opening up this PR. I'm having trouble understanding what was broken. Could you please provide an instance where the current code doesn't work?

rmgk:
The comment of the changed line does provide an example

current: 
```
❯ echo "[1]: |value|"  | string replace --regex ": \|(.*)\|" ' \$1'
[1] $1
```

changed
```
❯ echo "[1]: |value|"  | string replace --regex ": \|(.*)\|" ' $1'
[1] value
```

so the current behavior just always prints $1 instead of the extracted value of the variable when searching the current environment

at least, that is the case for me on fish version 3.3.1

PatrickF1:
I see what you're saying. But your example doesn't work for me:
```
echo "[1]: |value|"  | string replace --regex ": \|(.*)\|" ' \$1'
[1] value
```
And this is on fish 3.3.1

PatrickF1:
I'm asking in the fish shell gitter https://gitter.im/fish-shell/fish-shell?at=619ecf3a026dc63ca614ce26

rmgk:
I found why it sometimes works for me: The old version does work without the "regex-easyesc" feature (current default, but will change at some point).
https://fishshell.com/docs/current/language.html#future-feature-flags

But the new version should (and for me does) work in both cases 🤔

PatrickF1:
So you're saying because `regex-easyesc` will eventually be on by default, we should merge this?

rmgk:
Yes, see the link for the details, the regex-easyescape feature was introduced in 3.1 (off by default) to fix essentially the behavior you ask about in the gitter. The documentation states that these opt in features become the default at some point.

PatrickF1:
Nice work! Thank you for finding this bug helping problems in the future! I will merge this.

rmgk:
Awesome! And thanks to you for maintaining this tool :-)

PatrickF1 (approved):

PatrickF1:
You're welcome! If you'd like to show additional appreciation, would you mind starring this repo please?

PatrickF1:
Btw, how did you reach https://fishshell.com/docs/current/language.html#future-feature-flags when you were investigating the bug?

rmgk:
When investigating why the issue did not appear immediately after deleting the fish_variables files, I realized that I do `set -U fish_features 3.0 3.1` which seemed like the obvious culprit for the different behavior. And then search for `fish shell fish_features` to get documentation.
As how I found out about fish_features in the first place … if I recall correctly I had some issues because the fish shell included in Ubuntu is quite old and I was reading the changelog to see the difference with newer versions where it is mentioned …

PR #211, opened 2021-11-29
Hide ./ prefix when fd version >= 8.3.0

kidonng:
[fd 8.3.0](https://github.com/sharkdp/fd/releases/tag/v8.3.0) introduced a `--strip-cwd-prefix` flag to revert a new behavior: `./` prefix is shown when output is not a tty (terminal).

This is what you get in `_fzf_search_directory` without this patch:
截屏2021-11-29 20 06 42
And with this patch:
截屏2021-11-29 20 06 53

### DIFF
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -6,7 +6,7 @@ function _fzf_search_directory --description "Search the current directory. Repl
         # Remove this logic March '22 iff fd docs document --strip-cwd-prefix
         set --prepend fd_opts --strip-cwd-prefix
     end
-    
+
     set fzf_arguments --multi --ansi $fzf_dir_opts
     set token (commandline --current-token)
     # expand any variables or leading tilde (~) in the token
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -1,5 +1,12 @@
 function _fzf_search_directory --description "Search the current directory. Replace the current token with the selected file paths."
     set fd_opts --color=always $fzf_fd_opts
+    if test (fd --version | string replace --regex --all '[^\d]' '') -ge 830
+        # fd 8.3 prepends ./ to all paths when output is being piped
+        # we don't need this so we hide it by passing --strip-cwd-prefix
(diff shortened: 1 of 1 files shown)


kidonng:
BTW, there are some typos on https://github.com/PatrickF1/fzf.fish/wiki/Thanks 😄

PatrickF1:
Thanks, I was just about to do this myself.

Btw, you have my email--feel free to email me the typos and I'll fix them. Sorry about that.

functions/_fzf_search_directory.fish
```diff
@@ -7,6 +7,12 @@ function _fzf_search_directory --description "Search the current directory. Repl
     # unescape token because it's already quoted so backslashes will mess up the path
     set unescaped_exp_token (string unescape -- $expanded_token)
 
+    # Hide ./ prefix when fd version >= 8.3.0
+    # https://github.com/sharkdp/fd/releases/tag/v8.3.0
```
PatrickF1, line 11: Could you move this up to line 3?
And could you add a todo here that we'll remove this check in a few months and `--strip-cwd-prefix` is included in fd's documentation (it's not currently there) and bump up the required fd version to 8.3.0?

PatrickF1 (approved):

PatrickF1:
Oh, actually could you add a test for when fd is > 8.3 and when it's < 8.3?

kidonng:
> Oh, actually could you add a test for when fd is > 8.3 and when it's < 8.3?

...using the actual program or a mocked fd? Either way it will be too much for this simple check. I don't like test for the test's sake.

PatrickF1:
Oh a second thought, I kind of like the idea of leaving in the ./ always and instead removing the logic that appends / to the path if it's a directory. Advantages
- less code
- now users can just hit enter to cd or execute if it's an executable
- works with both pre 8.3 and post 8.3

kidonng:
No, it will look ugly with a base directory (`base/./path`). Also note that fd won't show the prefix it _is_ connected to a tty, because of the same reason: the prefix does not make sense to human eyes.

PatrickF1:
You're right--it is pretty ugly. Ok we'll stick with this version!

PR #47, opened 2020-10-17
Illustrate features using gifs instead of screenshots

patrickf3139:
Goal
- Use gifs--a video is much better at capturing and explaining a feature than a static image.
- Feature @IlanCosman's Tide prompt manager

For posterity, here is how I'm recording
- remove all Nerd font glyphs from prompt as they don't render with asciinema
- disable fish history with `set -g fish_history ""` to turn off annoying autocomplete
- iTerm2 settings: dimensions = 121x33, font size = 16
- record using `asciinema`, output file outside of git repo so git section of prompt doesn't change
- convert to gif at https://dstein64.github.io/gifcast/
- gifcast settings: font = courier, theme = solarized dark

Resolves #31

PR #216, opened 2021-12-30
Introduce feature to search for processes

cr-mitmit:
Allows searching for PIDs of any running process.
Useful for commands such as

> kill -STOP \co 

Todos
- [X] make the fzf options for pid configurable https://github.com/PatrickF1/fzf.fish#pass-fzf-options-for-a-specific-feature
- [X] ~add process owner to main fzf window: CANCELLED not enough space~
- [X] add memory and CPU use to preview window
- [x] remove columns that are very rarely used from preview window
- [x] make sure everything is OS-portable (checked Mac OS, debian, and Ubuntu manpages)
- [x] feature in readme
- [ ] add a gif to readme
- [X] add a smoke test
- [X] consider setting key to ctrl+option+p because it's more mnemonic (which is advertised in the readme, so we better follow it)
- [x] vsz vs rss vs %mem
- [ ] use -o format to make names cleaner in preview


PatrickF1:
Hi @cr-mitmit ! I love how thorough you were with this PR, even updating the printed help and the readme!

Unfortunately, and I'll be upfront, this one is gonna be tricky!
The first problem is that there is no preview. Normally, I wouldn't require preview, but you do need the preview to figure out which process you want to work with.
And the second problem that you'll run into with generating the preview, and even right now, is that the ps command options varies widely between OSes and I haven't figured out which options are reliable.

cr-mitmit:
Hi.

Re preview, the flags I use are `ps -Aa -opid,command`. They seem to documented for many years on the linux side, and work and tested on Macos Monterey. This seems to be as portable as it gets.

The default table shows the pid and the full command line (including arguments).
We could generate a preview with more information, but I'm also unsure what else to show. I could theoretically add parent pid to the table as well.
The current implementation already returns only the pid upon selection. 
Thanks for the pointer to the search git log, I'll see if there's better way than an awk call.  

Would love your thoughts

PatrickF1:
> Re preview, the flags I use are ps -Aa -opid,command. They seem to documented for many years on the linux side, and work and tested on Macos Monterey. This seems to be as portable as it gets.

Great! Let's keep the cross-OS compatibility in mind. Also, let's be sure to include Cygwin.

> We could generate a preview with more information, but I'm also unsure what else to show. I could theoretically add parent pid to the table as well.

Yeah definitely being able to see the parent pid and command would be helpful. Maybe start time as well? TBH I'm not sure. That's why I'd like to explore this topic more by re-asking this question: "may I ask which use cases you see this being useful for?" Personally, I've never had to kill a process using the commandline so I have no intuition as to what users will need.

cr-mitmit:
Hi.
Will do. I don't have access to a Cygwin machine and documentation has conflicting views  on whether the `ps` command in Cygwin respects or ignores `-o` formatting. The one in WSL, ironically, does.

I'll look into adding the preview including the parent information. 
As for use-cases, I sometimes find myself doing trying to kill a process / service. I liked the idea of using fzf to search for the process. Basically the interactive equivalent of `kill (pgrep XXX)`

cr-mitmit:
Also added a preview. As far as I can tell uses only portable flags across both Linux, WSL and Macos

functions/_fzf_search_pids.fish
```diff
@@ -0,0 +1,19 @@
+function _fzf_search_pids --description "Search pid for all running commands"
+    set fzf_arguments --multi --ansi $fzf_pids_opts
+    set token (commandline --current-token)
+    # expand any variables or leading tilde (~) in the token
+    #set expanded_token (eval echo -- $token)
+
+    set pid_selected (
+        ps -A -opid,command | \
+        _fzf_wrapper --query $token  \
+                     --preview='ps -f -p {1}' | \
```
PatrickF1, line 10: Sorry I was unclear. I didn't mean an explicit fzf preview, though that's nice. I mean it'd be good to show everything that the user needs in fzf directly so that it can be skimmed and filtered on by fzf's search.

PatrickF1:
> Will do. I don't have access to a Cygwin machine and documentation has conflicting views on whether the ps command in Cygwin respects or ignores -o formatting. The one in WSL, ironically, does.

Hmm that's tough. @kidonng  sorry to bother you. Do you have Windows experience to help out here?

functions/_fzf_search_pids.fish
cr-mitmit: I actually think it's less useful that way...
Or rather, the current implementation simply puts the pid and command line (and relegates the other details for the preview), I think just adding columns such as parent pid are less useful.
What are your thoughts?

PatrickF1: You're right--those things aren't that useful for searching and would clutter the view. I'm thinking the user will narrow down the process to a few matching the name and then use the preview to find the exact process to operate on. In that case, since the preview window offers tons of space and it's not expensive to put more info there, can you find a few more fields that might be of interest to the user?

PatrickF1: looking at how I've been using activity monitor, maybe memory use, CPU use, IO use?

PatrickF1: Hmm I see that idea is complicated by this: https://github.com/dalance/procs#permissions-issues. Do you see a need to be able to kill processes not owned by the current user? If not, I wonder if we should not show processes owned by other users.

cr-mitmit: Not sure we can find good portable flags for that. 
Generally, I've taken the view of using the `-f` flag which offers a sane default. 
Not sure what we'd want to add there.

As for processes by the user only, I thought about it, but at least in my use-case I quite often do something like `sudo kill XXX`. 

One option would be to check the full command and see if it begins with sudo, or if the euid is root and change behavior based on that. WDYT?

cr-mitmit: Or maybe use commandline --current-process to extract the last token and see if it begins with sudo. 
I fear it's a bit fragile...

PatrickF1: > Generally, I've taken the view of using the -f flag which offers a sane default.

Ok that seems fair.

> One option would be to check the full command and see if it begins with sudo, or if the euid is root and change behavior based on that. WDYT?

I dont' want to get into the business of parsing the user's command. And if we're not going to try displaying memory and disk use, then there's no point in doing it.

functions/_fzf_search_pids.fish
```diff
@@ -0,0 +1,19 @@
+function _fzf_search_pids --description "Search pid for all running commands"
+    set fzf_arguments --multi --ansi $fzf_pids_opts
+    set token (commandline --current-token)
+    # expand any variables or leading tilde (~) in the token
+    #set expanded_token (eval echo -- $token)
+
+    set pid_selected (
+        ps -A -opid,command | \
```
PatrickF1, line 8: We need to filter out PID COMMAND headers from the output sent into fzf

cr-mitmit: Added to the commit (--header-lines=1)

functions/_fzf_search_pids.fish
cr-mitmit: Cool. Dropping

PatrickF1:
Nice! I think we're very close. The last thing I'd like to see if if we can show the username instead of uid in the preview. Is that possible and portable? Reason: I noticed that Mac's Activity Monitor shows username in all of its windows and so it seems to be a pretty important thing to know, and users don't want to have to deal with ids.

functions/_fzf_configure_bindings_help.fish
```diff
@@ -12,6 +12,7 @@ DESCRIPTION
         Search git status  |  Ctrl+Alt+S (S for status)    |  --git_status
         Search history     |  Ctrl+R     (R for reverse)   |  --history
         Search variables   |  Ctrl+V     (V for variable)  |  --variables
+        Search processes   |  Ctrl+O     (O for prOcess)   |  --pids
```
PatrickF1, line 15: this says 0 but you got it bought to o. which one do you think is best?

cr-mitmit: Might it be a font issue? I've verified again all letters are O (capital o) and not Zero
The binding is done to \co to mimic the way Ctrl-R is bound to \cr (and since Ctrl binding are case insensitive).
Therefore, Ctrl-O is bound to the Process list.

Am I missing something?

cr-mitmit:
Also added the -O user to add the user-name to the process list. Good idea

functions/_fzf_configure_bindings_help.fish
PatrickF1: Oops my bad it is a capital o indeed. Ok I think that's fine...
Btw can you make the option --processes instead of --pids?

PatrickF1:
> Also added the -O user to add the user-name to the process list. Good idea

Can you put -O user so it comes first in the output?

cr-mitmit:
Changed the command to user processes instead of pids (Not just the doc, across the commit).

As for where the user listing appears, the only way to specific the order explicitly is to forgo the `-f` flag in favor of explicitly listing a set of options. While it *should* work, I'm not sure how portable that would be. I think the value isn't worth the risk in this case.

PatrickF1:
> As for where the user listing appears, the only way to specific the order explicitly is to forgo the -f flag in favor of explicitly listing a set of options. While it should work, I'm not sure how portable that would be. I think the value isn't worth the risk in this case.

Good point. I will take it upon myself to figure out the set of portable options and figure out the best one to use from here. As the plugin author, I need to worry about having a sane set of defaults and portability so I'll need to do this anyway. Please wait patiently as I have a lot of real work to catch up on returning from vacation, but rest assured I should get this in within a week. Thanks for opening up this PR and your hard work so far!

PatrickF1:
> Changed the command to user processes instead of pids (Not just the doc, across the commit).

Also I think you might've forgotten to push this commit.

cr-mitmit:
> > Changed the command to user processes instead of pids (Not just the doc, across the commit).
> 
> Also I think you might've forgotten to push this commit.

Pushed. Sorry

PatrickF1:
Todos for myself (feel free to chime in with your feedback)
- make the fzf options for pid configurable https://github.com/PatrickF1/fzf.fish#pass-fzf-options-for-a-specific-feature
- add process owner to main fzf window
- add memory and CPU use to preview window
- remove columns that are very rarely used from preview window
- make sure everything is OS-portable (but first figure out which OSes fish supports)
- add a gif to readme

cr-mitmit:
1. I had actually added a fzf_processes_opts but then forgot to actually pass it to fzf_wrapper 🤦‍♂️
2. Should be easy, add user to the options  -opid,user,command. It should be portable enough across Mac, Linux, and WSL. Can't tell re Cygwin as it uses a weird version of PS
3. We can likely achieve it with a set of flags to -o. Portability worries me here though

Sounds great. 
If I have time today I'll try to add a patch that does exactly that.

cr-mitmit:
I've implemented a new preview, but getting memory into M and G (as opposed to KB which is what PS returns) is ugly as hell.
Any thoughts on how to make this preview bearable?
Technically, it sets the PS, converts the RSS number into strings with M and G depending on the memory usage and the painfully recreates a table and accounting for processes whose name contains a space (so all "columns" after the 6th are actually fake because of spaces in the command line).

It's ugly as hell... but works...

>                     --preview='ps -o pid,user,%cpu,time,rss,command -p {1} | awk \'NR>1 {if ($5 > 1048576) {$5=sprintf("%.02fG", $5/1024/1024);} else {$5=sprintf("%.02fM",$5/1024);}}{for(i=1;i<=5;i++){printf("%s|", $i); $i=""}; $0=$0; print; }\' | column -s"|" -t' \

PatrickF1:
What if you just returned the memory and CPU as percentages? It’s already
built in

cr-mitmit:
I find it less usable.
My Mac has 64G of ram, knowing my browser is taking 5% is just depressing, but not as useful.

I'll push the change that shows pid, user, %cpu, ram, command line with the awk based printing.

Let me know what you think.

Also, I experimented with adding the user to the main table, but found that it takes too much screen real-estate.
Love to hear your thoughts.

PatrickF1:
> I find it less usable. My Mac has 64G of ram, knowing my browser is taking 5% is just depressing, but not as useful.

I hear you that it's not as useful. However, when it comes down to slight usability vs ease of maintenance / reducing dependencies, I'm going to pick ease of maintenance / reducing dependencies. Not having to write a complicated awk line and not requiring users to have a decently recent version of awk installed is much better.

> Also, I experimented with adding the user to the main table, but found that it takes too much screen real-estate.
Love to hear your thoughts.

You're right--I just tried it and if I'm using my laptop screen as opposed to monitor or I have my terminal 1/2 screen then we lose the ability to read the command. Let's keep the user column out. Thanks for the feedback. And thanks for collaborating with me!

cr-mitmit:
> I hear you that it's not as useful. However, when it comes down to slight usability vs ease of maintenance / reducing dependencies, I'm going to pick ease of maintenance / reducing dependencies. Not having to write a complicated awk line and not requiring users to have a decently recent version of awk installed is much better.

If that's the case, let's remove formatting and leave it at the rss number. This means the unit will be the default in multiples of 1024B (i.e. number of KB in ram).

That would make it very simple -opid,user,%cpu,rss,command and we can remove all the complex awk scripts.

WDYT?

PatrickF1:
That's a good compromise! Thanks for thinking not just about what you need
but what works best for the broad majority of fzf.fish users!

PatrickF1:
Btw, I realized the OSes we need to support are simply found on https://fishshell.com/ under "Go Fish". I'll add some checkboxes to the title post to make sure we cover all of them when we check for portability of our ps usage.

cr-mitmit:
> That's a good compromise! Thanks for thinking not just about what you need but what works best for the broad majority of fzf.fish users!

The commit I just pushed removes all fancy formatting and leaves the RSS field as is.
Definitely cleaner code without fancy awk scripting...

functions/_fzf_search_processes.fish
```diff
@@ -0,0 +1,21 @@
+function _fzf_search_processes --description "Search pid for all running commands"
```
PatrickF1, line 1: Search all running processes. Replace the current token with the id of the selected process."

functions/_fzf_search_processes.fish
```diff
@@ -0,0 +1,21 @@
...
+    set fzf_arguments --multi --ansi $fzf_processes_opts
```
PatrickF1, line 2: Can you document $fzf_processes_opt on the readme please?

functions/_fzf_search_processes.fish
```diff
@@ -0,0 +1,21 @@
...
+    set token (commandline --current-token)
+    # expand any variables or leading tilde (~) in the token
```
PatrickF1, line 4: I think you can delete this comment.

functions/_fzf_search_processes.fish
```diff
@@ -0,0 +1,21 @@
...
+    #set expanded_token (eval echo -- $token)
+
+    set pid_selected (
+        ps -A -opid,command | \
+        _fzf_wrapper --query $token  \
+                     --header-lines=1 \
+                     --preview='ps -o pid,user,%cpu,time,rss,command -p {1}' \
+                     $fzf_arguments | \
+        awk '{print $1}'
```
PatrickF1, line 13: how come you still need awk?

functions/_fzf_search_processes.fish
```diff
@@ -0,0 +1,21 @@
...
+    set fzf_arguments --multi --ansi $fzf_processes_opts
+    set token (commandline --current-token)
+    # expand any variables or leading tilde (~) in the token
+    #set expanded_token (eval echo -- $token)
+
+    set pid_selected (
+        ps -A -opid,command | \
+        _fzf_wrapper --query $token  \
+                     --header-lines=1 \
+                     --preview='ps -o pid,user,%cpu,time,rss,command -p {1}' \
```
PatrickF1, line 11: could you also add start time (`start`) to the fields previewed?

functions/_fzf_search_processes.fish
```diff
@@ -0,0 +1,21 @@
+function _fzf_search_processes --description "Search pid for all running commands"
+    set fzf_arguments --multi --ansi $fzf_processes_opts
+    set token (commandline --current-token)
+    # expand any variables or leading tilde (~) in the token
+    #set expanded_token (eval echo -- $token)
+
+    set pid_selected (
+        ps -A -opid,command | \
```
PatrickF1, line 8: Check this out from the ps manpage
```
-c      Change the “command” column output to just contain the executable name, rather than the full command line.
```
should we use it?

functions/_fzf_search_processes.fish
cr-mitmit: Good point. The output from FZF is the entire line, but I'm trying to extract just the PID here to be useful for commands such as "kill". the `awk` here is simply extracting the pid from the selected line

functions/_fzf_search_processes.fish
cr-mitmit: I intentionally didn't use it. I think being able to list and filter based on the command line arguments might be usable. 
Putting just the command will leave us with a lot of pretty opaque processes for things like shell scripts, ssh sessions etc....

PatrickF1: good point! I agree. Your ssh sessions argument convinced me.

functions/_fzf_search_processes.fish
PatrickF1: Ah so we want to reduce dependencies so you could just use standard fish functions to do this. Check out how I did it in `_fzf_search_git_log`

PatrickF1:
Don't worry I haven't forgotten about this! I will try to take another pass at it today. We are close!

cr-mitmit:
Hi. Any news?

PatrickF1:
Hi @cr-mitmit. I am _so_ sorry. I'm honestly pretty behind on life but this is still important to me and still high up on my to do list (except other things keep getting added on top if it as fast as I can complete them). If you'd like to speed it up, you could try to tackle one of both of my last todos:
- add a gif to readme (see https://github.com/PatrickF1/fzf.fish/pull/47 for how)
- add a test that makes sure the ps command succeeds (don't need to check the output sent into fzf, this is purely for compatibility testing in case we change the ps command)

functions/_fzf_search_processes.fish
```diff
@@ -0,0 +1,19 @@
...
+        _fzf_wrapper --query $token  \
+                     --header-lines=1 \
+                     --preview='ps -o pid,user,%cpu,rss,time,start,command -p {1}' \
+                     $fzf_arguments
+    )
+
+    if test $status -eq 0
+        set pid_selected (string split -n " " $process_selected)[1]
+        commandline --current-token --replace -- (string escape -- $pid_selected)
```
PatrickF1, line 15: found a bug: this works for one process but not multiple. probably should be doing `$process_selected)[1]`

PatrickF1: also, is string escape necessary?

PatrickF1:
Hey I decided to switch to using %mem despite the argument you made earlier. rss and vsz are too confusing and do not match with Mac's Activity Monitor's . Plus we can change this easily without worrying about backwards compatibility since this is only used in the preview

PatrickF1:
Ok sorry for the delay. I decided to just merge this in so you don't have to wait any longer and just implement the other changes I want by myself. Sorry it took so long--besides being very busy, I am also a bit of a perfectionist.

cr-mitmit:
Thank you - for the merge as well as writing fzf.fish!

PatrickF1:
You're welcome! I'm spurred on by everyone who contributes and uses it! :)
Btw, now that you've got contributed such a major feature to fzf.fish, would you mind starring it? A personal goal of mine is to reach 1k stars.

PR #228, opened 2022-02-28
[Search processes] account for exited processes in preview

kidonng:
First, thanks for your effort in getting the process search feature in! I saw the mention but unfortunately don't have the time to chime in.

I was not satisfied as you did, and made some local changes to fzf.fish. It was until #227 got merged that I realized I never submitted those patches (yes, I have been putting the preview window on the bottom for the longest time).

So here is the change that I really want to upstream: mark exited process as red.

Screen Shot 2022-02-28 at 19 59 59

Since [the preview is provided by a second `ps`](https://github.com/PatrickF1/fzf.fish/blob/a033d17b6d0288400b34c69911822e879b7d98bd/functions/_fzf_search_processes.fish#L12), the selected process may have already exited and all we get is a confusing header.

_Where is that process I want to look at?_

Even if it exited, the full process name along with its parameter is still useful, but without a functioning preview, the only chance you see it in whole is to have a super-duper wide display (does this even exist?)

No more. The process info is right in fzf itself and we just need to instruct fzf to use _that_ instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Use `fzf.fish` to interactively find and insert the shell entities listed below 
 
 - **Search input:** the pid and command of all running processes, outputted by `ps`
 - **Key binding and mnemonic:** <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>P</kbd> (`P` for process)
-- **Preview window:** the CPU usage, memory usage, start time, and other information about the process (shown in red if exited)
+- **Preview window:** the CPU usage, memory usage, start time, and other information about the process
 
 _The prompt used in the screencasts was created using [IlanCosman/tide](https://github.com/IlanCosman/tide)._
 
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Use `fzf.fish` to interactively find and insert the shell entities listed below 
 - **Search input:** all the variable names of the environment currently [in scope][var scope]
 - **Key binding and mnemonic:** <kbd>Ctrl</kbd>+<kbd>V</kbd> (`V` for variable)
 - **Preview window:** the scope info and values of the variable
-- `$history` is excluded for technical reasons so use the search command history feature instead to inspect it
+- `$history` is excluded for technical reasons so use the [search command history](#a-previously-run-command) feature instead to inspect it
 
 ### Process ids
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,15 +56,15 @@ Use `fzf.fish` to interactively find and insert the shell entities listed below
 - **Search input:** all the variable names of the environment currently [in scope][var scope]
 - **Key binding and mnemonic:** <kbd>Ctrl</kbd>+<kbd>V</kbd> (`V` for variable)
 - **Preview window:** the scope info and values of the variable
-- `$history` is excluded for technical reasons so use the search command history feature instead to inspect it
+- `$history` is excluded for technical reasons so use the [search command history](#a-previously-run-command) feature instead to inspect it
```
kidonng, line 59: Irrelevant but this is a small improvement I would like to see.

functions/_fzf_search_processes.fish
```diff
@@ -9,7 +9,8 @@ function _fzf_search_processes --description "Search all running processes. Repl
                      --ansi \
                      # first line outputted by ps is a header, so we need to mark it as so
                      --header-lines=1 \
-                     --preview="ps -o '$ps_preview_fmt' -p {1}" \
+                     # 40 is the length from "PID" to "COMMAND" in the header "PID PARENT USER  %CPU RSS_IN_KB START_TIME COMMAND"
```
kidonng, line 12: This spacing may not be portable across different implementations of `ps` but we are not going to jump in the rabbit hole and check `ps` variant, like in the original implementation PR 🤷

functions/_fzf_search_processes.fish
```diff
@@ -9,7 +9,8 @@ function _fzf_search_processes --description "Search all running processes. Repl
...
+                     --preview="ps -o '$ps_preview_fmt' -p {1} || begin; set_color red; echo {1}(string repeat --count 40 ' '){2..}; end" \
```
PatrickF1, line 13: can you make this a helper function that fish autoloads? similar to _fzf_extract_var_info

README.md
PatrickF1: nice!

PatrickF1:
Just to check my understanding: what this change does is if the process can no longer be found by ps, take the info that is already stored in the fzf results and just echo that in red?

PatrickF1:
@kidonng still interested in working on this?

PatrickF1:
Hi @kidonng, I decided to go for nice-and-simple: just print an error message if the process has already exited. I want to reduce complexity as much as possible, and also I think the user can easily look to the left to find the pid and command. It's not perfect because sometimes the command is too long, but I think it's worth it.

PR #230, opened 2022-03-08
[Search git status] fix extra quotes from spaced file paths

### LINKED ISSUE #220: search git search status doesn’t work for files with spaces
## Describe the bug

The file paths outputted by git status, when there is a space in the path, are not valid. They contain an extra layer of quotes.

## Steps to reproduce

touch "a b.txt"
select "a b.txt" in search git status
notice the output is `'"a b.txt"'`, and if you try to cat it you'll get `cat: "a b.txt": No such file or directory`

## Comments
Looking for contributors to fix this one! @kidonng if you're interested 😉
The test `tests/search_git_status/outputs_right_paths.fish` will also need to be updated. It attempts to test paths with spaces but it's not catching the bug :(

PatrickF1:
This fixes a bug where file paths outputted by search git status include an extra layer of single quotes if the path includes a space. Turns out this is because we string escape all our paths before writing them to the command line but git status already string escapes paths with spaces.

Fixes #220

### DIFF
--- a/functions/_fzf_search_git_status.fish
+++ b/functions/_fzf_search_git_status.fish
@@ -25,7 +25,7 @@ function _fzf_search_git_status --description "Search the output of git status. 
                 end
             end
 
-            commandline --current-token --replace -- (string escape -- $cleaned_paths | string join ' ')
+            commandline --current-token --replace -- (string join ' ' $cleaned_paths)
         end
     end
 
--- a/tests/search_git_status/outputs_right_paths.fish
+++ b/tests/search_git_status/outputs_right_paths.fish
@@ -1,10 +1,11 @@
-set files "filename_with_*.txt" "filename with space.csv"
+set files "filename_with_*.txt" "filename with space.csv" tests/_resources/nestedfilename
 touch $files
 mock commandline \* ""
 mock commandline "--current-token --replace --" "echo \$argv"
 set --export --append FZF_DEFAULT_OPTS "--filter='filename'"
+
 set actual (_fzf_search_git_status)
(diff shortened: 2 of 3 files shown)


tests/wrapper/doesnt_clobber_env.fish
```diff
@@ -1,5 +1,6 @@
 set --global --export SHELL /bin/sh
-# don't set FZF_DEFAULT_OPTS so it will set one
+# erase FZF_DEFAULT_OPTS so that one will be set by the wrapper function
+set --erase FZF_DEFAULT_OPTS
```
PatrickF1, line 3: out of scope change to facilitate local tests where I do have a FZF_DEFAULT_OPTS set already

functions/_fzf_search_git_status.fish
```diff
@@ -25,7 +25,7 @@ function _fzf_search_git_status --description "Search the output of git status.
                 end
             end
 
-            commandline --current-token --replace -- (string escape -- $cleaned_paths | string join ' ')
```
PatrickF1, line 28: Turns out this is unnecessary because git status already escapes filenames with spaces in them

tests/search_git_status/outputs_right_paths.fish
```diff
@@ -1,10 +1,11 @@
-set files "filename_with_*.txt" "filename with space.csv"
+set files "filename_with_*.txt" "filename with space.csv" tests/_resources/nestedfilename
```
PatrickF1, line 1: updated test to also test it works for nested filenames

PR #237, opened 2022-05-10
[Search git status] preview changed files

NextAlone:
Add a preview for the changed file

### DIFF
--- a/functions/_fzf_preview_changed_file.fish
+++ b/functions/_fzf_preview_changed_file.fish
@@ -1,4 +1,4 @@
-function _fzf_preview_status
+function _fzf_preview_changed_file
     set -l path (string split ' ' $argv)[-1]
     if string match -r '^\?\?' $argv >/dev/null
         echo Untracked
--- a/functions/_fzf_preview_changed_file.fish
+++ b/functions/_fzf_preview_changed_file.fish
@@ -1,21 +1,21 @@
 function _fzf_preview_changed_file
     set -l path (string split ' ' $argv)[-1]
-    if string match -r '^\?\?' $argv >/dev/null
+    if string match -r '^\?\?' $argv --quiet
         echo Untracked
         echo
         _fzf_preview_file $path
-    else if string match -r '^A\s\s\S' $argv >/dev/null
+    else if string match -r '^A\s\s\S' $argv --quiet
         echo -e (set_color green)New file
         echo
         git diff --color=always -- /dev/null $path
     else
-        if string match -r '\S\s\S' $argv >/dev/null
+        if string match -r '\S\s\S' $argv --quiet
             echo -e (set_color red)Unstaged
             echo
             git diff --color=always -- $path
             echo
         end
-        if string match -r '\S\s\s\S|\S\S\s\S' $argv >/dev/null
+        if string match -r '\S\s\s\S|\S\S\s\S' $argv --quiet
             echo -e (set_color green)Staged
             echo
             git diff --color=always --cached HEAD -- $path
--- a/functions/_fzf_preview_changed_file.fish
+++ b/functions/_fzf_preview_changed_file.fish
@@ -4,10 +4,6 @@ function _fzf_preview_changed_file
         echo Untracked
         echo
         _fzf_preview_file $path
-    else if string match -r '^A\s\s\S' $argv --quiet
-        echo -e (set_color green)New file
-        echo
-        git diff --color=always -- /dev/null $path
     else
         if string match -r '\S\s\S' $argv --quiet
             echo -e (set_color red)Unstaged
@@ -15,7 +11,7 @@ function _fzf_preview_changed_file
             git diff --color=always -- $path
             echo
         end
-        if string match -r '\S\s\s\S|\S\S\s\S' $argv --quiet
+        if string match -r '^\S' $argv --quiet
             echo -e (set_color green)Staged
             echo
             git diff --color=always --cached HEAD -- $path
--- a/functions/_fzf_preview_changed_file.fish
+++ b/functions/_fzf_preview_changed_file.fish
@@ -1,5 +1,5 @@
 # helper for _fzf_search_git_status
-# input should be a line from git status --short, something like:
+# arg should be a line from git status --short, something like...
 # MM functions/_fzf_preview_changed_file.fish
 function _fzf_preview_changed_file
     set -l path (string split ' ' $argv)[-1]
@@ -10,7 +10,6 @@ function _fzf_preview_changed_file
         if string match -r '\S\s\S' $argv --quiet
             echo -e (set_color --underline red)=== Unstaged ===\n
             git diff --color=always -- $path
-            echo
         end
         if string match -r '^\S' $argv --quiet
             echo -e (set_color --underline green)=== Staged ===\n
--- a/functions/_fzf_preview_changed_file.fish
+++ b/functions/_fzf_preview_changed_file.fish
@@ -1,20 +1,20 @@
+# helper for _fzf_search_git_status
+# input should be a line from git status --short, something like:
+# MM functions/_fzf_preview_changed_file.fish
 function _fzf_preview_changed_file
     set -l path (string split ' ' $argv)[-1]
     if string match -r '^\?\?' $argv --quiet
-        echo Untracked
-        echo
+        echo -e (set_color --underline)=== Untracked ===\n
         _fzf_preview_file $path
     else
         if string match -r '\S\s\S' $argv --quiet
-            echo -e (set_color red)Unstaged
-            echo
+            echo -e (set_color --underline red)=== Unstaged ===\n
             git diff --color=always -- $path
             echo
         end
         if string match -r '^\S' $argv --quiet
-            echo -e (set_color green)Staged
-            echo
-            git diff --color=always --cached HEAD -- $path
+            echo -e (set_color --underline green)=== Staged ===\n
+            git diff --color=always --staged -- $path
         end
     end
 end
--- a/functions/_fzf_preview_status.fish
+++ b/functions/_fzf_preview_status.fish
@@ -1,19 +1,21 @@
 function _fzf_preview_status
-    echo $argv
     set -l path (string split ' ' $argv)[-1]
-    echo $path
     if string match -r '^\?\?' $argv >/dev/null
+        echo Untracked
+        echo
         _fzf_preview_file $path
-    else if string match -r '^A' $argv >/dev/null
+    else if string match -r '^A\s\s\S' $argv >/dev/null
+        echo -e (set_color green)New file
+        echo
         git diff --color=always -- /dev/null $path
     else
-        if string match -r '[MDTRC] \S' $argv >/dev/null
+        if string match -r '\S\s\S' $argv >/dev/null
             echo -e (set_color red)Unstaged
             echo
             git diff --color=always -- $path
             echo
         end
-        if string match -r '[AMDTRC][MDTRC ]\S' $argv >/dev/null
+        if string match -r '\S\s\s\S|\S\S\s\S' $argv >/dev/null
             echo -e (set_color green)Staged
             echo
             git diff --color=always --cached HEAD -- $path
--- a/functions/_fzf_preview_status.fish
+++ b/functions/_fzf_preview_status.fish
@@ -1,0 +1,22 @@
+function _fzf_preview_status
+    echo $argv
+    set -l path (string split ' ' $argv)[-1]
+    echo $path
+    if string match -r '^\?\?' $argv >/dev/null
+        _fzf_preview_file $path
+    else if string match -r '^A' $argv >/dev/null
+        git diff --color=always -- /dev/null $path
+    else
+        if string match -r '[MDTRC] \S' $argv >/dev/null
+            echo -e (set_color red)Unstaged
+            echo
+            git diff --color=always -- $path
+            echo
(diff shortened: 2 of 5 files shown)


PatrickF1:
Don't let the fact the other two PRs didn't get merged discourage you. It was more because the authors lost interest than this feature being impossible to work on (I hope).
If you can make the preview show both the staged and working directory diffs against HEAD, then I am happy to merge.
Two potential solutions:
- run git diff and git diff —index separately and append them to a single file, then cat it
https://stackoverflow.com/questions/67150991/show-both-worktree-and-index-changes-for-a-single-path-in-git
- use git status vv

NextAlone:
> Don't let the fact the other two PRs didn't get merged discourage you. It was more because the authors lost interest than this feature being impossible to work on (I hope).
> If you can make the preview show both the staged and working directory diffs against HEAD, then I am happy to merge.
> Two potential solutions:
> - run git diff and git diff —index separately and append them to a single file, then cat it
> https://stackoverflow.com/questions/67150991/show-both-worktree-and-index-changes-for-a-single-path-in-git
> - use git status vv

Ok, i will check these two.

NextAlone:
> Don't let the fact the other two PRs didn't get merged discourage you. It was more because the authors lost interest than this feature being impossible to work on (I hope). If you can make the preview show both the staged and working directory diffs against HEAD, then I am happy to merge. Two potential solutions:
> 
> * run git diff and git diff —index separately and append them to a single file, then cat it
>   https://stackoverflow.com/questions/67150991/show-both-worktree-and-index-changes-for-a-single-path-in-git
> * use git status vv

Are you talking about being displayed together or separately?

NextAlone:
How about this?

```fish
if string match -r \'\?\?\' {1} > /dev/null
    _fzf_preview_file {2..}
else if string match -r \'A\' {1} > /dev/null
    git diff --color=always -- /dev/null {2..}
else
    git diff --color=always HEAD {2..}
end'
```

PatrickF1:
> Are you talking about being displayed together or separately?

If there are both staged and working tree changes, I want the preview to show both.

Yeah the second one is along the lines of what I'm thinking. I guess you don't need the intermediate file to store them huh?

NextAlone:
```
 X          Y     Meaning
 -------------------------------------------------
          [AMD]   not updated
 M        [ MTD]  updated in index
 T        [ MTD]  type changed in index
 A        [ MTD]  added to index
 D                deleted from index
 R        [ MTD]  renamed in index
 C        [ MTD]  copied in index
 [MTARC]          index and work tree matches
 [ MTARC]    M    work tree changed since index
 [ MTARC]    T    type changed in work tree since index
 [ MTARC]    D    deleted in work tree
             R    renamed in work tree
             C    copied in work tree
 -------------------------------------------------
 D           D    unmerged, both deleted
 A           U    unmerged, added by us
 U           D    unmerged, deleted by them
 U           A    unmerged, added by them
 D           U    unmerged, deleted by us
 A           A    unmerged, both added
 U           U    unmerged, both modified
 -------------------------------------------------
 ?           ?    untracked
 !           !    ignored
 -------------------------------------------------
```
Not works with R(renamed), C(copied), T(type changed), and unmerged status. i will find out the way.

NextAlone:
> > > Are you talking about being displayed together or separately?
> > 
> > 
> > If there are both staged and working tree changes, I want the preview to show both.
> > Yeah the second one is along the lines of what I'm thinking. I guess you don't need the intermediate file to store them huh?
> 
> Yes, no intermediate file, and add check with status (just simply string match, don't know other convenient way)?
> 
> ```fish
> if string match -r '\?\?' {1} > /dev/null
>     _fzf_preview_file {2..}
> else if string match -r 'A' {1} > /dev/null
>     git diff --color=always -- /dev/null {2..}
> else 
>     if string match -r '\S\s\S' > /dev/null
>         echo -e (set_color red)Unstaged
>         echo
>         git diff --color=always HEAD -- {2..}
>         echo
>     end
>     if string match -r '\s\s|\S\S\s' > /dev/null
>         echo (set_color green)Staged
>         echo
>         git diff --color=always --cached HEAD -- {2..}
>     end
> end
> ```
> 
> image

maybe this is enough to preview

functions/_fzf_search_git_status.fish
```diff
@@ -8,6 +8,7 @@ function _fzf_search_git_status --description "Search the output of git status.
             _fzf_wrapper --ansi \
                 --multi \
                 --query=(commandline --current-token) \
+                --preview='set -l path (string split " " {})[-1]; if not string match -r "\?\?.*" {}; if string match -r "^[NMAD]\s\s\S*" {}; git diff --color=always --cached -- $path; else if string match -r "^\s[NMAD]\s\S*" {}; git diff --color=always -- $path; else if string match -r "^[NMAD][NMAD]\s\S*" {}; git diff --color=always -- $path; git diff --color=always --cached -- $path; end; end' \
```
PatrickF1, line 11: Can you make this a helper function so it's easier to review and maintain? see the preview function for other features for examples

functions/_fzf_preview_status.fish
```diff
@@ -0,0 +1,24 @@
+function _fzf_preview_status
```
PatrickF1, line 1: can you renamed to changed file?

functions/_fzf_preview_status.fish
```diff
@@ -0,0 +1,24 @@
...
+    set -l path (string split ' ' $argv)[-1]
+    if string match -r '^\?\?' $argv >/dev/null
```
PatrickF1, line 3: having a bit of trouble following the logic of each if/else if/else. could you outline the "algorithm" at a high level and explain why you think it's comprehensive?

PatrickF1 (commented):
thanks for the great work so far!

functions/_fzf_preview_status.fish
NextAlone: you mean `_fzf_preview_changed_file` ?

functions/_fzf_preview_status.fish
NextAlone: 1. use `_fzf_preview_file` to preview untracked files or dirs
2. diff with /dev/null and new files
3. then other files are indexed, then seperate unstaged and staged files as you mentioned.

I think that this basically covers common situations

functions/_fzf_preview_status.fish
PatrickF1: yeah, sorry I wasn't clear

functions/_fzf_preview_changed_file.fish
```diff
@@ -0,0 +1,24 @@
+function _fzf_preview_changed_file
+    set -l path (string split ' ' $argv)[-1]
+    if string match -r '^\?\?' $argv >/dev/null
+        echo Untracked
+        echo
+        _fzf_preview_file $path
```
PatrickF1, line 6: niceeeee, so clever to go call _fzf_preview_file
can you go to the top of _fzf_preview_file and remove the comment?

functions/_fzf_preview_changed_file.fish
```diff
@@ -0,0 +1,24 @@
+function _fzf_preview_changed_file
+    set -l path (string split ' ' $argv)[-1]
+    if string match -r '^\?\?' $argv >/dev/null
```
PatrickF1, line 3: you can do --quiet to avoid >/dev/null

functions/_fzf_preview_changed_file.fish
```diff
@@ -0,0 +1,24 @@
...
+        echo Untracked
+        echo
+        _fzf_preview_file $path
+    else if string match -r '^A\s\s\S' $argv >/dev/null
```
PatrickF1, line 7: For the regex, I get why the ^A, but how come you also have to do \s\s\S?

functions/_fzf_preview_changed_file.fish
```diff
@@ -0,0 +1,24 @@
...
+        echo
+        git diff --color=always -- /dev/null $path
+    else
+        if string match -r '\S\s\S' $argv >/dev/null
+            echo -e (set_color red)Unstaged
+            echo
+            git diff --color=always -- $path
+            echo
+        end
+        if string match -r '\S\s\s\S|\S\S\s\S' $argv >/dev/null
```
PatrickF1, line 18: sorry I know you tried to explain what you were doing but I'm still confused about this regex and the one above

functions/_fzf_preview_changed_file.fish
NextAlone: to match `[AMDCRT][MCDRT] file`

I haven't tried to simplify it since it works.

functions/_fzf_preview_changed_file.fish
NextAlone: Maybe `^\S` also works.

functions/_fzf_preview_changed_file.fish
NextAlone: is this? 
`# helper function for _fzf_search_directory`

NextAlone:
No need to split `^A`, it works with staged file

functions/_fzf_preview_changed_file.fish
PatrickF1: yup, actually can you update it to say `helper function for _fzf_search_directory and _fzf_preview_changed_file`

NextAlone: > yup, actually can you update it to say `helper function for _fzf_search_directory and _fzf_preview_changed_file`

emm, you mean `_fzf_search_directory` and `_fzf_search_git_status` right?

NextAlone:
> Thanks for the updates. This week is really busy so it'll be a while before I can come back to review but just letting you know I won't forget about this

It doesn't matter, I will always be here.

NextAlone:
> Thanks for the updates. This week is really busy so it'll be a while before I can come back to review but just letting you know I won't forget about this

It doesn't matter, I will always be here.

PatrickF1:
Don't worry I haven't forgotten about this, just thinking through what is the best way to make this feature simple and easy to use!

functions/_fzf_preview_changed_file.fish
```diff
@@ -0,0 +1,20 @@
+function _fzf_preview_changed_file
+    set -l path (string split ' ' $argv)[-1]
+    if string match -r '^\?\?' $argv --quiet
```
PatrickF1, line 3: Hi, I'm back! I've found a great way to extract the symbols without using regex--you can use `git status --porcelain`. Can you update the code to use that instead? Afterwards, we'll be very close to merging

NextAlone: I didn't quite understand it and didn't find its use effect

image

PatrickF1: Try using version 1 porcelain and qualifying the file path. The status it displays is much easier to parse because it's at the very beginning of the line. Then you can use substring instead of playing with unreadable regexes.

NextAlone: > 

but it still needs to regex with prefix to detect file status

PatrickF1: What I'd strongly prefer to see is something like this pattern
```fish
set file_status (command to grab the two letters of the file status)
if test file_status[0] = "A"

```
I think this'll be much more readable. Right now I still don't exactly know what your regexes are matching.

NextAlone: > What I'd strongly prefer to see is something like this pattern
> 
> ```fish
> set file_status (command to grab the two letters of the file status)
> if test file_status[0] = "A"
> ```
> 
> I think this'll be much more readable. Right now I still don't exactly know what your regexes are matching.

`--porcelain` no diff to `-s`
image

NextAlone: > What I'd strongly prefer to see is something like this pattern
> 
> ```fish
> set file_status (command to grab the two letters of the file status)
> if test file_status[0] = "A"
> ```
> 
> I think this'll be much more readable. Right now I still don't exactly know what your regexes are matching.

`git status` shows with two character, when first not null, it means staged changes, while second, the unstaged.

So, `\S\s\S` means unstaged and `^\S` means staged

PatrickF1: Got it, I get it now, thanks! I am fine with the logic as it is then.

NextAlone:
And if there any other changes request？

PatrickF1:
Nope, I'll take it from here and merge soon :)

PatrickF1:
Thanks for working on this and especially our patience @NextAlone! Really appreciate it. You are the third person to try to PR this change in and it seems third time's the charm. Sorry I've been so busy.

NextAlone:
> Thanks for working on this and especially our patience @NextAlone! Really appreciate it. You are the third person to try to PR this change in and it seems third time's the charm. Sorry I've been so busy.

Also thanks for your work, this is such a wonderful plugin.

NextAlone:
> Thanks for working on this and especially our patience @NextAlone! Really appreciate it. You are the third person to try to PR this change in and it seems third time's the charm. Sorry I've been so busy.

Also thanks for your work, this is such a wonderful plugin.

davidnanry:
Thank you both! This is an excellent addition to an excellent plugin

NextAlone:
@PatrickF1 And it is time to update `Modified paths` section in Readme.

PatrickF1:
I did update it. I'll add a gif later. Or did you mean something else?

PR #238, opened 2022-05-10
[Search git log] include diff stats in preview

NextAlone:


PatrickF1:
Could you include a screenshot and explain why you think this should be added?

NextAlone:
> Could you include a screenshot and explain why you think this should be added?

If one commit is huge, an overview on the top maybe helpful to quickly get which files were changed.
image

PatrickF1:
Did you test this on smaller width screens? I see --stat has a width parameter and i wonder if we should limit it

NextAlone:
> Did you test this on smaller width screens? I see --stat has a width parameter and i wonder if we should limit it
It works good, i think width is not needed
image

PatrickF1 (approved):
Ok great! Thanks for testing it for me.

PatrickF1:
Oh wait how come you also added `--oneline`?

PatrickF1:
Oh I see, is it because you are running two git show commands, one on top of the other, and the first one is just for the diff so you want a short summary of the commit?

PatrickF1:
Ok I suspect there's a way to run git show to show both the stats and the diff itself. I'll look into it if you'd like.

NextAlone:
> Oh I see, is it because you are running two git show commands, one on top of the other, and the first one is just for the diff so you want a short summary of the commit?

Yes, you are right.

NextAlone:
> Ok I suspect there's a way to run git show to show both the stats and the diff itself. I'll look into it if you'd like.

I tried looking for it, but couldn't find it. It would be nice if there was a way to show them together.

PatrickF1:
git show HEAD --stat --patch seems to work. Can you try it out?

NextAlone:
> git show HEAD --stat --patch seems to work. Can you try it out?

It works, and --patch-with-stat the same.

image

PatrickF1:
Thank you @NextAlone for this great contribution!

NextAlone:
> Thank you @NextAlone for this great contribution!

And thanks for your work too.

PR #239, opened 2022-05-11
[Search variables] Always wrap preview

jaminthorns:


PatrickF1:
Hi, can you explain why?

jaminthorns:
My bad! I should've written up an explanation for this PR.

My `FZF_DEFAULT_OPTS` very purposefully doesn't include `--preview-window='wrap'`, as I generally find wrapped lines confusing when viewing files and diffs. That means that for the functions that `fzf.fish` defines, the preview is not going to wrap unless it's set for that command (either explicitly in the function or through one of the `_opts` variables).

For searching history and processes, we use `--preview-window="bottom:<LINES>:wrap"` (3 lines for the command, [4 lines for the process info](https://github.com/PatrickF1/fzf.fish/pull/227)). This is a sensible default because the preview content is likely going to be a single line or few lines (commands and process info) that could be long (which is why we wrap). I figure variable values are a similar situation, and I actually ran into this while previewing a database URL variable that was cut off.

Now that I explain this, though, do you think the preview window for searching variables should be the same as for searching history (`--preview-window="bottom:3:wrap"`)?

PatrickF1:
Hmm I see. Do you mind sharing the variables where you need the wrapping? TBH in my daily use I have never encountered variables that are too long width or height wise to need wrapping or to know whether the window should be on the right or the bottom

PatrickF1 (approved):

PR #245, opened 2022-06-22
[Search history] don't merge history when in private mode

kidonng:
Calling `history merge` in private mode (`fish --private`) **clears** the session history. So if you execute the search history function while in private mode, there is no history to search.

### DIFF
--- a/functions/_fzf_search_history.fish
+++ b/functions/_fzf_search_history.fish
@@ -1,5 +1,6 @@
 function _fzf_search_history --description "Search command history. Replace the command line with the selected command."
     # history merge incorporates history changes from other fish sessions
+    # skip merging in private mode as it clears current session history
     if test "$fish_private_mode" != 1
         builtin history merge
     end
--- a/functions/_fzf_search_history.fish
+++ b/functions/_fzf_search_history.fish
@@ -1,7 +1,7 @@
 function _fzf_search_history --description "Search command history. Replace the command line with the selected command."
     # history merge incorporates history changes from other fish sessions
     # skip merging in private mode as it clears current session history
-    if test "$fish_private_mode" != 1
+    if test -n "$fish_provate_mode"
         builtin history merge
     end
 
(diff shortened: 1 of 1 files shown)


functions/_fzf_search_history.fish
```diff
@@ -1,6 +1,9 @@
 function _fzf_search_history --description "Search command history. Replace the command line with the selected command."
     # history merge incorporates history changes from other fish sessions
-    builtin history merge
+    # skip merging in private mode as it clears current session history
+    if test "$fish_private_mode" != 1
```
IlanCosman, line 4: This should be `test -n "$fish_private_mode"`

IlanCosman (changes requested):

PatrickF1:
> Calling history merge in private mode (fish --private) clears the session history

Are you saying it clears the session history of the current session but leaves the other sessions' history intact, or it clears every session's history?

PatrickF1:
This seems like a bug that should be fixed in fish. Why would they cause it to clear history?

kidonng:
But we can fix it now, instead of waiting for identifying if this is really an issue and fish release a new version with fix.

PatrickF1:
Sure! In the meantime, I've filed a bug report https://github.com/fish-shell/fish-shell/issues/9050

PatrickF1:
Oh wow bug already got fixed in fish-shell master. Looks like the fix is to make it error out so regardless, it was good that we  merged this.

functions/_fzf_search_history.fish
```diff
@@ -1,6 +1,9 @@
 function _fzf_search_history --description "Search command history. Replace the command line with the selected command."
     # history merge incorporates history changes from other fish sessions
-    builtin history merge
+    # skip merging in private mode as it clears current session history
+    if test -n "$fish_provate_mode"
```
PatrickF1, line 4: oops just realized there's a typo here. I'll fix it

kidonng: Time to get a spell checker for vim I guess 😅

Issue #251, opened 2022-08-03
multi select for history (ctrl+r) puts timestamps in front of all but the first command


acegallagher:
## Before proceeding...

- [ X] My CLIs meet the [minimum version requirements](https://github.com/PatrickF1/fzf.fish#installation)
- [ X] I have read through the [troubleshooting guide](https://github.com/PatrickF1/fzf.fish/wiki/Troubleshooting)
- [ X] If I updated fzf.fish for the first time in over 6 months, I read the [migration guides](https://github.com/PatrickF1/fzf.fish/wiki/Migration-Guides) to make sure I'm caught up

## Describe the bug

ctrl+r and then tab to select multiple entries puts the timestamp in front of the entries, like:

![selecting commands](https://user-images.githubusercontent.com/25115064/182724823-ef06460d-a986-4516-afca-8040edc535e0.jpeg)
![after selected](https://user-images.githubusercontent.com/25115064/182724825-d9af41cc-6eac-48b2-9643-86e89c3274ba.jpeg)

I imagine this is repeatable behavior for the package, but I'm happy to help debug of course if it's somehow only my setup. I'm also looking at the code now and trying to see if I could submit a pull request.

## Environment

Versions installed:
bat 0.21.0 (405e5f74)
fd 8.4.0
fzf 0.31.0 (a0ef898)
fish, version 3.4.1

- terminal: konsole
- OS: Linux 5.18.14

Thanks much for putting together something so well thought out.

acegallagher:
Ah yeah, I have a pull request, just needed to change a few lines. Here is the code snippet:

`        set -U commands_selected # empty list to loop over selected commands, in case multiple                                        
        for c in (string split \n $command_with_ts)                                                                                   
            set --append commands_selected (string split --max 1 " │ " $c)[2]                                                         
        end                                                                                                                           
        commandline --replace -- $commands_selected  `

Incoming!

PatrickF1:
Hi @acegallagher, the search history feature does not multi-select by default, so you must've configured it to be multi-selectable. What's your use case for selecting multiple commands?

PR #252, opened 2022-08-03
[Search history] support selecting multiple commands

acegallagher:
Resolves https://github.com/PatrickF1/fzf.fish/issues/251.
Reply the issue I opened, here's the minor change. It's a big deal for my workflow to be able to cobble together scripts this way. 

Hopefully it's helpful! Thanks

### DIFF
--- a/functions/_fzf_search_history.fish
+++ b/functions/_fzf_search_history.fish
@@ -20,9 +20,11 @@ function _fzf_search_history --description "Search command history. Replace the 
     )
 
     if test $status -eq 0
-        set commands_selected # empty list to loop over selected commands, in case multiple
+        set commands_selected
+        set ts_capture_regex '^(\d\d-\d\d \d\d:\d\d:\d\d │ )'
+        # remove timestamps from commands before putting them into command line
         for c in (string split \n $commands_with_ts)
-            set --append commands_selected (string replace --regex '^(\d\d-\d\d \d\d:\d\d:\d\d\s│\s)' '' $c)
+            set --append commands_selected (string replace --regex $ts_capture_regex '' $c)
         end
         commandline --replace -- $commands_selected
     end
--- a/functions/_fzf_search_history.fish
+++ b/functions/_fzf_search_history.fish
@@ -19,9 +19,13 @@ function _fzf_search_history --description "Search command history. Replace the 
     )
 
     if test $status -eq 0
-        set -U commands_selected # empty list to loop over selected commands, in case multiple
-        for c in $commands_with_ts
-            set --append commands_selected (string split --max 1 " │ " $c)[2]
+        set commands_selected # empty list to loop over selected commands, in case multiple
+        for c in (string split \n $commands_with_ts)
+            if string match -r --quiet '(\d\d-\d\d \d\d:\d\d:\d\d)' $c
+                set --append commands_selected (string split --max 1 " │ " $c)[2]
+            else
+                set --append commands_selected $c
+            end
         end
         commandline --replace -- $commands_selected
     end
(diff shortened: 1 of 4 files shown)


PatrickF1:
Gotcha, that makes sense. I can see myself--though rare--selecting multiple commands to run. Okay I'll work on this.

functions/_fzf_search_history.fish
```diff
@@ -19,8 +19,11 @@ function _fzf_search_history --description "Search command history. Replace the
     )
 
     if test $status -eq 0
-        set command_selected (string split --max 1 " │ " $command_with_ts)[2]
-        commandline --replace -- $command_selected
+        set -U commands_selected # empty list to loop over selected commands, in case multiple
+        for c in (string split \n $command_with_ts)
```
PatrickF1, line 23: Commands can be multi-line. That's why the test broke. Thankfully, fish variables are arrays so you can just iterate through them. Can you try
```
for c in $command_with_ts)
```
Also can you rename the variable to `commands_with_ts`?

PatrickF1 (changes requested):

acegallagher:
There you go, requested changes made and committed

PatrickF1:
Thank you! So...a complication is that this doesn't work when you select a multi line command. I'll look into a solution when I have a significant contiguous chunk of free time some time later. But feel free to suggest ideas or fix it yourself.

PatrickF1:
Btw how do you comment and commit using two different GH accounts?

acegallagher:
Ohhhhhhhhhhhh. There will have to be some sort of regex that only removes the date if present. Hilariously, the multi-account was unintentional. It's because I made the commit with an ssh key shared by both accounts (changed the key for one of them without thinking about it). Oops.

acegallagher:
Alright, there's a solution. It will strip the date from any line the date exists by regex. I suppose there's potentially some conflict if there's a line that also has the exact format of the date but I have no idea how much that would matter to you. It wouldn't to me.

functions/_fzf_search_history.fish
```diff
@@ -19,8 +19,15 @@ function _fzf_search_history --description "Search command history. Replace the
     )
 
     if test $status -eq 0
-        set command_selected (string split --max 1 " │ " $command_with_ts)[2]
-        commandline --replace -- $command_selected
+        set commands_selected # empty list to loop over selected commands, in case multiple
+        for c in (string split \n $commands_with_ts)
+            if string match -r --quiet '(\d\d-\d\d \d\d:\d\d:\d\d)' $c
```
PatrickF1, line 24: Great idea! I think this can all be simplified if you just did `string replace --regex` and replaced all instances of that string. I'd also add a `^` so it only matches dates in the beginning of the line.

functions/_fzf_search_history.fish
```diff
@@ -5,7 +5,7 @@ function _fzf_search_history --description "Search command history. Replace the
         builtin history merge
     end
 
-    set command_with_ts (
+    set commands_with_ts (
         # Reference https://devhints.io/strftime to understand strftime format symbols
         builtin history --null --show-time="%m-%d %H:%M:%S │ " |
         _fzf_wrapper --read0 \
```
PatrickF1, line 11: could you add `--multi` now that we're officially supporting it, please?

acegallagher:
The regex now only matches the beginning of the string and no need for the if statement. Much cleaner. Also added multi!

Thanks <3

PatrickF1:
Thanks for putting in the work, Michael. I'll take it from here--just gotta update the readme and add a test and I'll merge!

acegallagher:
No problem. Congratulations on having a readable well built project that's very useful. Will continue looking through the code and imagining other ways to contribute.

PatrickF1 (approved):

PatrickF1:
Great work Michael! Thank you for your mind words. Would you mind starring my repo? It's a goal of mine to reach 1k stars. :)

PR #258, opened 2022-08-28
Add a "grep directory" keybinding

sartak:
This PR adds a new kind of search: grep directory. It's similar to "search directory" but instead of fuzzy-matching file _names_, grep directory does regex-matching on file _content_, powered by `rg`. The default hotkey is ctrl-g.

Here's what it looks like, and one example of why I think it's cool.

grep-directories

There's a few minor shortcomings that I left out for now to avoid complicating the code. If you think this feature is worthy of inclusion, I'd be happy to give any or all of these a shot!

* This hardcodes `rg`, and so there's no way to switch to `ag`, or `grep -r`.
* If you hit the ctrl-g hotkey with an incomplete token in the buffer, we unconditionally pass that along to `rg` as the initial regex. We could be a little more nuanced. fzf.fish's directory search checks if the input ends with a slash and is a directory, and if so, sets that as the base directory for `fzf`'s input. I think that'd be a good fit here too.
* Running `rg` with no regular expression shows a list of empty results. We can probably do something better than that, but I'm not sure what. (Maybe default to showing all lines when user input is empty, to better match the "filtering" affordance)
* ~The preview window just shows the top of the file. We should automatically scroll to the matching line.~ _edit: Implemented!_
* The preview window should highlight the matching line. To do so we need to thread `--highlight-line {2}` from `_fzf_grep_directory` into `bat`. The blocker here is `_fzf_preview_file`'s comment of "we collect all the arguments into one single [path] variable". (I temporarily hacked in support for highlighting the matching line in the preview video above, but it's not shippable.)

Anyway, thanks for fzf.fish! Let me know if there's anything you'd like me to improve on here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,9 @@ First, install a proper version of these CLI dependencies:
 | [fzf][]  | 0.27.2                   | fuzzy finder that powers this plugin           |
 | [fd][]   | 8.3.0                    | faster and more colorful alternative to `find` |
 | [bat][]  | 0.16.0                   | smarter `cat` with syntax highlighting         |
+| [rg][]   | (optional)               | faster and more powerful alternative to `grep` |
 
-The search directory feature uses [fd][] and [bat][] to list and preview files. If your package manager [doesn't install them as `fd` and `bat`](https://github.com/PatrickF1/fzf.fish/wiki/Troubleshooting#search-directory-feature-does-not-work) respectively, then you can symlink them to those names.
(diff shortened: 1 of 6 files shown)


PatrickF1:
Hi Shawn, thanks for being interested in contributing. To be honest, this is a really tough feature to add for various reasons. See my previous thoughts on it here. https://github.com/PatrickF1/fzf.fish/pull/118#issuecomment-780738223

Even if we implement this exactly the way I specified in the comment, I'm still not sure I would merge it.

sartak:
That's unfortunate :/ Thanks again for fzf.fish.

PatrickF1:
Sorry about the wasted time spent on this. And thanks for being a good sport and understanding!

sartak:
It’s all good! Most of the time was spent on the implementation, which I can still keep for myself :)

PR #262, opened 2022-09-26
[Search git status] support previewing merge conflict

PatrickF1:
Also, update some comments and docs for Search git status.

### DIFF
--- a/tests/preview_changed_file/deleted_in_working.fish
+++ b/tests/preview_changed_file/deleted_in_working.fish
@@ -1,5 +1,5 @@
 mock git diff "echo \$argv"
 set output (_fzf_preview_changed_file " D out.log")
 
-contains -- "| Unstaged |" $output && not contains -- "| Staged |" $output
+contains "| Unstaged |" $output && not contains "| Staged |" $output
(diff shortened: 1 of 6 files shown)


tests/preview_changed_file/deleted_in_working.fish
```diff
@@ -1,7 +1,5 @@
 mock git diff "echo \$argv"
-set file dir/file.txt
```
PatrickF1, line 2: $file was only being used in one place

tests/preview_changed_file/deleted_in_working.fish
```diff
@@ -1,7 +1,5 @@
...
+set output (_fzf_preview_changed_file " D out.log")
 
-set output (_fzf_preview_changed_file " D $file")
-
-contains -- "| Unstaged |" $output && not contains "| Staged |" $output
```
PatrickF1, line 6: add -- to the second invocation of `contains` to protect it from any output starting with `-`

PatrickF1, line 6: contains doesn't need `--` because by passing in the first argument, it would never interpret a line prefixed with`-` in $output as an option.

PR #280, opened 2022-12-07
[Search Directory] Ensure unaliased fd is called

### LINKED ISSUE #281: aliasing fd causes Search Directory to hang
I have fd aliased to a less pager, like so:

```fish
function fd --wraps=fdfind --description 'alias fd=fdfind'
  if isatty stdout
    command fdfind --color always $argv | less
  else
    command fdfind $argv
  end
end
```

This is causing Directory Search to hang until all results have been found.

#280 Fixes than by checking that unwrapped `fd` is called.

Susensio:
If fd is aliased, Search Directory will hang until fd finishes executing. This is because fish functions buffer. Resolves #281.

### DIFF
--- a/functions/_fzf_search_directory.fish
+++ b/functions/_fzf_search_directory.fish
@@ -33,7 +33,7 @@ function _fzf_search_directory --description "Search the current directory. Repl
         if test (count $file_paths_selected) = 1
             set commandline_tokens (commandline --tokenize)
             if test "$commandline_tokens" = "$token" -a -d "$file_paths_selected" \
-                    -a (fd --version | string replace --regex --all '[^\d]' '') -lt 840
+                    -a (command fd --version | string replace --regex --all '[^\d]' '') -lt 840
                 set file_paths_selected $file_paths_selected/
             end
         end


functions/_fzf_search_directory.fish
```diff
@@ -16,10 +16,10 @@ function _fzf_search_directory --description "Search the current directory. Repl
         set --append fd_opts --base-directory=$unescaped_exp_token
         # use the directory name as fzf's prompt to indicate the search is limited to that directory
         set --prepend fzf_arguments --prompt="Search Directory $unescaped_exp_token> " --preview="_fzf_preview_file $expanded_token{}"
-        set file_paths_selected $unescaped_exp_token(fd $fd_opts 2>/dev/null | _fzf_wrapper $fzf_arguments)
+        set file_paths_selected $unescaped_exp_token(command fd $fd_opts 2>/dev/null | _fzf_wrapper $fzf_arguments)
```
PatrickF1, line 19: Hi Susensio, unfortunately we should not always use the unwrapped version of fd because some users might actually aliased it to overcome an issue on Linux where fd is installed as fdfind. https://github.com/PatrickF1/fzf.fish/discussions/93

PatrickF1:
Re-opening based on further discussion in #281 .

Susensio:
I've fixed one dangling alias that I've missed. It wasn't blocking but for completeness it should be unaliased too.

I don't know why the test are failing.

PatrickF1:
The test is failing for the same reason that the test is failing in #282.

PatrickF1:
Closing in favor of #282. Next time, just update the same PR. 😉️

PR #284, opened 2022-12-27
feat: allow configuring FZF colors

gregorias:
fzf.nvim already relies on highlighting done by fd and bat, but not all colors are set by them. This setting allows people to configure FZF colors not affected by fd or bat while keeping default options fzf.nvim provides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,12 @@ fzf supports global default options via the [FZF_DEFAULT_OPTS](https://github.co
 
 `fzf.fish` locally sets [a sane `FZF_DEFAULT_OPTS` whenever it executes fzf](functions/_fzf_wrapper.fish). If you export your own `FZF_DEFAULT_OPTS`, then yours will be used instead.
 
+### Customize FZF colors
+
+To customize FZF colors, use `fzf_fish_colors` environment variable. If you do not set your own `FZF_DEFAULT_OPTS`, `fzf.fish` will set `$fzf_fish_colors` colors as FZF's colors. For example, `set -x fzf_fish_colors 'fg+:#665c54,bg+:#f2e5bc'`.
+
+[Fzf-color-picker](https://minsw.github.io/fzf-color-picker/) can help you pick color overrides you want.
+
 ### Pass fzf options for a specific command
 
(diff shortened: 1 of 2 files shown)


PatrickF1:
Hi Grzegorz, thanks for the PR. How come you prefer this over just allowing the user to set the colors in `FZF_DEFAULT_OPTS`?

gregorias:
> Hi Grzegorz, thanks for the PR. How come you prefer this over just allowing the user to set the colors in `FZF_DEFAULT_OPTS`?

I prefer this, because setting `FZF_DEFAULT_OPTS` overrides the defaults that fzf.fish provides. This means that to keep those defaults, I would need to copy them into my own config and hope they don't change. Since the only thing I need is to set colors, it's a more maintainable set up for me if there's a configuration knob for that.

PatrickF1:
> I prefer this, because setting FZF_DEFAULT_OPTS overrides the defaults that fzf.fish provides. This means that to keep those defaults, I would need to copy them into my own config and hope they don't change. Since the only thing I need is to set colors, it's a more maintainable set up for me if there's a configuration knob for that.

Ah that makes sense! I've long had on my todo list a way to set global fzf opts for all of fzf.fish and haven't had time to get to it. Would that resolve this need? I really don't want to make a fzf.fish flag that merely maps to a simple fzf flag.

PatrickF1:
> I would need to copy them into my own config and hope they don't change

gregorias:
> Ah that makes sense! I've long had on my todo list a way to set global fzf opts for all of fzf.fish and haven't had time to get to it. Would that resolve this need? I really don't want to make a fzf.fish flag that merely maps to a simple fzf flag.

Sounds reasonable to me. I'll just override FZF_DEFAULT_OPTS then. Feel free to close this PR.

Thanks.

PatrickF1:
Thanks for opening this issue and giving me insight into why people might not want to override FZF_DEFAULT_OPTS! When I get to it, I'll probably remove setting a default for it and just suggest it in the docs. It's definitely out of scope for the plugin to be messing with fzf globally.

PR #287, opened 2023-01-03
Added GPG output check

### LINKED ISSUE #286: Git log showing raw GPG output as commits
## Before proceeding...

- [X] My CLIs meet the [minimum version requirements](https://github.com/PatrickF1/fzf.fish#installation)
- [X] I have read through the [troubleshooting guide](https://github.com/PatrickF1/fzf.fish/wiki/Troubleshooting)
- [X] If I updated fzf.fish for the first time in over 6 months, I read the [migration guides](https://github.com/PatrickF1/fzf.fish/wiki/Migration-Guides) to make sure I'm caught up

## Describe the bug

If you happen to have GPG signed commits and your repo or global git configuration showw GPG signatures by default (setting `showSignature = true` on the repo or the global `.gitconfig`) , the signatures appear on the git log as entries that parse with errors. An example is this very repo, see the capture below:

image

with my global `.gitconfig` being like this:
```bash
[user]
        signingKey = xxxx
        name = xxx
        email = xxx
[commit]
        gpgSign = true
[core]
        editor = vim
[gpg]
        program = gpg
[log]
        showSignature = true 
[init]
        defaultBranch = master
```

If you force git to not show GPG signatures (by commenting showSignature = true), this of course disappears

image

but I see this suboptimal since it hides the signature information

## Steps to reproduce
Use git log on a repo with gpg-signed commits 

## Environment

Versions installed:

- Fish: 3.5.1
- fzf.fish: 9.5
- terminal: Konsole
- OS: Arch Linux

Which, if any, [configuration variables](https://github.com/PatrickF1/fzf.fish#configuration) such as `fzf_preview_file_cmd` are set?
None 

## Additional context
None

c4tich:
Fixed the raw GPG output that leaked in the commit list for repos that have the `showSignature = true` git configuration. The values of the check are specified in the [`%G?` format specifier](https://git-scm.com/docs/git-log#Documentation/git-log.txt-emGem) for `git log`

Resolves #286

### DIFF
--- a/functions/_fzf_search_git_log.fish
+++ b/functions/_fzf_search_git_log.fish
@@ -4,9 +4,9 @@ function _fzf_search_git_log --description "Search the output of git log and pre
     else
         # see documentation for git format placeholders at https://git-scm.com/docs/git-log#Documentation/git-log.txt-emnem
         # %h gives you the abbreviated commit hash, which is useful for saving screen space, but we will have to expand it later below
-        set log_fmt_str '%C(bold blue)%h%C(reset) - %C(cyan)%ad%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)'
+        set log_fmt_str '%C(bold blue)%h%C(reset) - [%G?] - %C(cyan)%ad%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)'
         set selected_log_lines (
-            git log --color=always --format=format:$log_fmt_str --date=short | \
+            git log --no-show-signature --color=always --format=format:$log_fmt_str --date=short | \
             _fzf_wrapper --ansi \
                 --multi \
                 --tiebreak=index \
--- a/functions/_fzf_search_git_log.fish
+++ b/functions/_fzf_search_git_log.fish
@@ -4,7 +4,7 @@ function _fzf_search_git_log --description "Search the output of git log and pre
     else
         # see documentation for git format placeholders at https://git-scm.com/docs/git-log#Documentation/git-log.txt-emnem
         # %h gives you the abbreviated commit hash, which is useful for saving screen space, but we will have to expand it later below
-        set log_fmt_str '%C(bold blue)%h%C(reset) - [%G?] - %C(cyan)%ad%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)'
+        set log_fmt_str '%C(bold blue)%h%C(reset) - %C(cyan)%ad%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)'
         set selected_log_lines (
             git log --no-show-signature --color=always --format=format:$log_fmt_str --date=short | \
             _fzf_wrapper --ansi \


functions/_fzf_search_git_log.fish
```diff
@@ -4,9 +4,9 @@ function _fzf_search_git_log --description "Search the output of git log and pre
     else
         # see documentation for git format placeholders at https://git-scm.com/docs/git-log#Documentation/git-log.txt-emnem
         # %h gives you the abbreviated commit hash, which is useful for saving screen space, but we will have to expand it later below
-        set log_fmt_str '%C(bold blue)%h%C(reset) - %C(cyan)%ad%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)'
+        set log_fmt_str '%C(bold blue)%h%C(reset) - [%G?] - %C(cyan)%ad%C(reset) %C(yellow)%d%C(reset) %C(normal)%s%C(reset)  %C(dim normal)[%an]%C(reset)'
```
PatrickF1, line 7: Let's leave this part out. I suspect most people won't care whether it was verified or not. I will add to my todo list making the log format string customizable so you can have this.

functions/_fzf_search_git_log.fish
```diff
@@ -4,9 +4,9 @@ function _fzf_search_git_log --description "Search the output of git log and pre
...
         set selected_log_lines (
-            git log --color=always --format=format:$log_fmt_str --date=short | \
+            git log --no-show-signature --color=always --format=format:$log_fmt_str --date=short | \
```
PatrickF1, line 9: This is the line that actually fixes the error message, right?

c4tich: Yes, this prevents the GPG raw output to leak

functions/_fzf_search_git_log.fish
c4tich: And we can't add like a parameter to configure if we want it to appear or not and default it to not appear? I agree most people won't care but I would like to have the option to keep it

PatrickF1: > I will add to my todo list making the log format string customizable so you can have this.

functions/_fzf_search_git_log.fish
PatrickF1: Can you explain why the %G still outputs despite `--no-show-signature`?

functions/_fzf_search_git_log.fish
c4tich: Wow it was just two sentences and I missed one of them. Sorry, I'll take the flag out

functions/_fzf_search_git_log.fish
c4tich: [`--show-signature`](https://git-scm.com/docs/git-log#Documentation/git-log.txt---show-signature) presents the output of `gpg --verify` and `--no-show-signature` negates this. I assume that both things are not incompatible, i.e., preventing the dump of an additional output is not incompatible with the format specifier

PatrickF1: Excuse my ignorance of GPG. What _does_ %G do then? How can it
> show "G" for a good (valid) signature, "B" for a bad signature, "U" for a good signature with unknown validity

without trying to do use GPG, which should result in errors if one is in a repo that doesn't use GPG signed commits or even have GPG installed?

c4tich: One way to understand it is by realizing that the flag [`--show-signature`](https://git-scm.com/docs/git-log#Documentation/git-log.txt---show-signature) is equivalent to the format specifier [`%GG`](https://git-scm.com/docs/git-log#Documentation/git-log.txt-emGGem):

> `%GG`
> raw verification message from GPG for a signed commit

> `--show-signature`
> Check the validity of a signed commit object by passing the signature to gpg --verify and show the output.

So, if we assume that `--show-signature` is equivalent to `%GG` it's immediate to understand that negating `--show-signature` (i.e. using `--no-show-signature`) is equivalent to removing `%GG` from the format string and thus eliminating the raw GPG output. And the fact that we remove one format specifier doesn't affect other format specifiers that are present.

Another way to see it is that `--show-signature` is shorthand of adding `%GG` to `log_fmt_str` and `--no-show-signature` is equivalent to removing it. Removing `%GG` doesn't affect other format specifiers defined in `log_fmt_str`. 

Hope this clears it a bit more

PatrickF1: Thanks that made a bit clearer. If you don't mind, I'm still a bit confused though.
If I add %G to the log format and use Search Git Log to view fzf.fish, everything is N even though some commits are  signed. I don't have gpg installed so was expecting a bunch of errors. What do you see when you use %G in fzf.fish?

c4tich: Hi @PatrickF1, I'm happy to be of assistance. 

First, I think you are aware of this, but I'll mention it anyways. The flag is '%G?' with the question mark at the end. You are probably using it that way because you get the expected output (N for a machine without GPG) but I see you keep referring to it as '%G', which doesn't exist. Just a nitpick

If I use `%G?`, in fzf.fish, I see this

withgpg


meaning that there are some signed commits whose signature couldn't be validated (because I haven't the public keys on this machine's keyring) but a signature was found nonetheless (i.e there's GPG). If there's no GPG installed, I indeed see a bunch of errors, and a bit wonky 

withoutgpg

I presume that someone that wishes to add the verification flag will have gpg on their path, but it would be nice to add some kind of warning or condition in the case that someone enables it without gpg. 

You say that you don't see errors. Could you please check the second line, below the commit hash on the right side of the window? If you look at the second image, you will see "`error: cannot run gpg: No such file or directory`". That should be your case too. In my case, the errors also flood the rest of the console in a messy way.

Let me know if you have any other doubts

PatrickF1: Thank you for replying, @c4tich! Yes, I meant `%G?`, and yeah I missed the error message `error: cannot run gpg`. That's my bad, I was too rushed. Thanks for coming back and explaining it all to me. It all makes sense to me now. 

I'll get on the git format log thing soon. It'll be a bit tricky though because fzf.fish makes assumptions on the structure of each line to parse the commit.

functions/_fzf_search_git_log.fish
PatrickF1: As promised! 039a86de01b43500bcc7af7b3a188e9093098cd1

PR #291, opened 2023-01-24
[Search Git Status] use box drawing chars for staged/unstaged header

yogan:
This makes the little header box for Git staged/unstaged/merged/etc. look a bit nicer.

Old:
```
+--------+
| Staged |
+--------+
```

New:
```
╭────────╮
│ Staged │
╰────────╯
```

It does not look perfect on GitHub web, but works in a terminal with a proper font:
image

We could make this optional (add some `$_fzf_fish_use_unicode_box` variable), but `fzf` itself is
also using those box drawing characters by default, so I guess pretty much everyone using `fzf.fish`
should be fine with it.

### DIFF
--- a/functions/_fzf_report_diff_type.fish
+++ b/functions/_fzf_report_diff_type.fish
@@ -1,16 +1,18 @@
 # helper for _fzf_preview_changed_file
 # prints out something like
-# +--------+
-# | Staged |
-# +--------+
+# ╭────────╮
+# │ Staged │
+# ╰────────╯
 function _fzf_report_diff_type --argument-names diff_type --description "Print a distinct colored header meant to preface a git patch."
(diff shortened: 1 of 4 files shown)


PatrickF1:
Hmm I forgot that it seems we are already using these chars here https://github.com/PatrickF1/fzf.fish/blob/main/functions/_fzf_search_history.fish#L11

yogan:
Hi Patrick! Yes, I have used the same characters that fzf does, and that's also how I "typed" them: by copy/paste from a fzf output. :-)
There are ways to type extended Unicode characters by some key combinations, but that depends on the OS, so I can't really give a clear answer for that. I usually just look them up when I need them. Wikipedia has a nice overview for example: https://en.wikipedia.org/wiki/Box-drawing_character

PR #292, opened 2023-01-31
[feat] Added "Search Units" function for searching systemd units

magnuslarsen:
I've added a command (CTRL+ALT+U - - U for systemd [U]nits)

I'm not too sure it fits the criteria of being widely useful, but as an admin, being able to search&select large swaps of services is super handy. I wanted to share in any case :-)

It's very simple, and is based on the Search Processes function; searching through `systemctl list-units` with common options (searching only services by default), and extensible via the `fzf_systemctl_opts`-variable  (i.e. if you want more than just services), returning only the selected unit names.

The preview window shows the status (with colors) of the selected unit, and if access is given (being apart of the `systemd-journal` or `adm` groups), it will show you the logs for that unit (systemctl has a default of 10 lines)

I have tested this with systemctl versions `237` (the current one on Ubuntu 18.04), `245` (current on Ubuntu 20.04), and `249` (current on Ubuntu 22.04)

I don't know if there is a fish equivalent to `head` which I've used. But that would be a nice improvement :-)

image

### DIFF
--- a/functions/_fzf_configure_bindings_help.fish
+++ b/functions/_fzf_configure_bindings_help.fish
@@ -17,6 +17,7 @@ DESCRIPTION
         Search History     |  Ctrl+R     (R for reverse)   |  --history
         Search Processes   |  Ctrl+Alt+P (P for process)   |  --processes
         Search Variables   |  Ctrl+V     (V for variable)  |  --variables
+        Search Units       |  Ctrl+Alt+U (U for units)     |  --units
     Override a command's binding by specifying its corresponding option with the desired key
     sequence. Disable a command's binding by specifying its corresponding option with no value.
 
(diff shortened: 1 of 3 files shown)


PatrickF1:
Hi Magnus! Really appreciate you putting in the work for this PR. Unfortunately, I think this is too niche and plus, I don't use systemd myself so it'd be really hard for me to maintain. The good news is, it shouldn't be too much work for you to make this a custom function in your dotfiles that integrates with fzf.fish so you won't lose your work. Thanks for understanding. I really need to keep the scope of fzf.fish manageable.

magnuslarsen:
No problem - they already are apart of my dotfiles :+1:

PatrickF1:
Great to hear!

PR #293, opened 2023-02-14
A proper implementation of fzf-tab

oddlama:
This implements fzf-tab completion as robustly as reasonably possible with fish. All remaining bugs are directly related to bugs (or weird design decisions) in fish. Some edge cases are different from edge cases in fish's completion, but I did absolutely everything in my power to create the best possible implementation (that is still reasonably simple). As you already pointed out, existing plugins like https://github.com/jethrokuan/fzf were known to be very very buggy because they were implemented without watching out for correctness.

- Since fish-shell/fish-shell#3469 was implemented last year, we now have the opportunity to create a completion that is actually less buggy than fish's internal pager 😎. This implementation works with weird characters in filenames such as `\t` `\n` and others, where the internal completion would fail
- I've made sure to delimit everything using `\0` in any logic to prevent introducing new edge-cases
- It actually isn't that much code, 50% are just comments to make sure that everyone knows what edge-cases exist

From what I could gather from #190 and #217 this still seems to be a controversial topic here. I just felt like implementing it correctly wasn't too complicated and would actually be less buggy than the built-in completion. I'd love if you would reconsider having a look at this and then decide if it is worth to you or not. I think a lot more people would benefit from it when it can be integrated here as opposed to me creating a new plugin for it. It's fine if you decide not to merge it, I just hope someone might find it useful.

### Feature overview

Generally, i tried mimicking Aloxaf's fzf plugin for zsh as much as possible, since it is the most mature fzf tab completion I could find. The main problem here being that zsh is just infinitely more capable when it comes to completions because it returns a lot of completion context, type and other information in a predictable manner, which are not available (or even existing) in fish. So all remaining issues all come from issues with fish that the internal completion also has. I've outlined known issues at the end.

Feature-wise, this means:
- fzf opens when pressing <kbd>Tab ⇥</kbd> only if there are ambiguities or multiple selections
- The completion adds spaces in a smart way. Spaces are added after each completion (when multiple options are selected), while no space added when completing a directory, link to a directory or options like `--color=`
- Weird filenames (such as `touch --mistake; rm --mis<TAB>`) will terminate options with `--` before inserting the file completion
- The initial query in fzf is set to the longest common prefix of all completions to avoid masking any options. If paths are completed, the longest matching prefix directory is even stripped to make the completion more clear.
- `~`, `$HOME` and other variables are handled correctly without requiring expansion

This is a baseline implementation, so any advanced features or config options some user's would like are not included (e.g. path prefix completion without opening fzf). I think it is a reasonable starting point from where users can customize it themselves if needed. There's theoretically some room for customization (description color, using fzf's preview, ...) which are not considered at the moment.

### Known Issues with fish's completion

There are some things that the internal completion can't do which we cannot fix:

- fish-shell/fish-shell#5363: folders ending in `=` break fish's internal completion (It looks like an argument with a file to fish: `this_folder_feels_like_an_argument=/`)
- fish-shell/fish-shell#784
- fish-shell/fish-shell#1711
- Probably some others.

It's important to know that fish cannot give reliable context information on completions. Knowing whether a completion is a file or argument can only be determined via a heuristic. Effectively this means some features are currently not viable:
- Filter based on completion type
- Do recursive completions (sudo CMD, runuser CMD, ...) 
- Prevent wrong completions like `--nonexisting-argument=/file/will/be/completed`

---

I hope that with this PR we can give fzf-tab another chance.

**EDIT**: Seems like this PR violates your formatting. This is the first time I write fish, so I'd be glad if you could tell me how to make the code compliant :)

Oh and users coming from fzf-tab might want to additionally use `set fzf_complete_opts --cycle --reverse --height=50%` to get the tab completion the way they are used to

### DIFF
--- a/functions/_fzf_complete.fish
+++ b/functions/_fzf_complete.fish
@@ -1,0 +1,178 @@
+# External limitations:
+# - fish-shell/fish-shell#9577: folders ending in '=' break fish's internal path completion
+#   (It thinks that it has to complete from / because `a_folder=/` looks like an argument to fish.
+# - fish cannot give reliable context information on completions. Knowing whether a completion
+#   is a file or argument can only be determined via a heuristic.
+set --global _fzf_complete_min_description_offset 14
+function _fzf_complete --description "Shell completion using fzf"
+	# Produce a list of completions from which to choose
+	# and do nothing if there is nothing to select from
+	set -l cmd (commandline -co) (commandline -t)
+	set -l completions (complete --escape --do-complete "$cmd")
+	test (count $completions) -eq 0; and return
+
+	set -l results
+	set -l first_completion (printf "%s" $completions[1] | cut -f1 --zero-terminated | string split0)
+
+	# A prefix that will be prepended before each selected completion
+	set -l each_prefix ""
+
+	# Only skip fzf if there is a single completion and it starts with the expected completion prefix.
+	# Otherwise, the completion originates from a match in the description which the user might
+	# want to review first (e.g. man ima<TAB> might match `feh`, an image viewer)
+	if test (count $completions) -eq 1; and test (string sub -l (string length -- (commandline -ct)) -- $first_completion) = (commandline -ct)
+		# If there is only one option then we don't need fzf
+		# Everytime we need to cut, we need zero termination to
+		# avoid newlines from being interpreted later.
+		set results $first_completion
+	else
+		# We preprocess the whole list to prevent spawning a slow cut process for each completion.
+		# From here on, make sure to use -- and zero termination everywhere to prevent introducing
+		# new edge-cases where characters are interpreted wrongly
+		set -l actual_completions (string join0 -- $completions | cut -f1 --zero-terminated | string split0)
+		set -l descriptions (string join0 -- $completions | cut -f2- --zero-terminated | string split0)
+
+		# Find the common prefix of all completions. Yes this seems to be a hard
+		# thing to do in fish if it should be fast. No external process may be
+		# spawned as this is too slow. So fish internals it is :/
+		set -l common_prefix $completions[1]
+		set -l common_prefix_length (string length -- $common_prefix)
+		for comp in $actual_completions[2..]
+			if test (string sub -l $common_prefix_length -- $comp) != $common_prefix
+				set -l new_common_prefix_length 0
+				set -l try_common_prefix_length (math min $common_prefix_length,(string length -- $comp))
+				# Binary search for new common prefix
+				set -l step $try_common_prefix_length
+				while test $step != 0
+					set step (math --scale 0 $step / 2)
+					# Adjust range to the left or right depending on whether the current new prefix matches
+					set -l xs (string sub -l $try_common_prefix_length -- $comp $common_prefix)
+					set -l op (test $xs[1] = $xs[2]; and printf "+"; and set new_common_prefix_length $try_common_prefix_length; or printf "-")
+					set try_common_prefix_length (math --scale 0 $try_common_prefix_length $op (math max 1,$step))
+				end
+				set common_prefix_length $new_common_prefix_length
+				set common_prefix (string sub -l $new_common_prefix_length -- $common_prefix)
+				# Stop if there is no common prefix
+				test $common_prefix_length = 0; and break
+			end
+		end
+
+		# If the common prefix includes a / we are completing a file path.
+		# Strip the prefix until the last / completely and later re-add it on the replaced token
+		set -l path_prefix (string match --regex --groups-only -- '^(.*/)[^/]*$' $common_prefix)
+		if test $status = 0
+			set -l path_prefix_length (string length -- $path_prefix)
+			set -l new_start (math 1 + $path_prefix_length)
+			for i in (seq (count $completions))
+				set completions[$i] (string sub -s $new_start -- $completions[$i])
+				set actual_completions[$i] (string sub -s $new_start -- $actual_completions[$i])
+			end
+
+			# We have a path-like prefix and will therefore strip this common prefix from all
+			# completions to un-clutter the menu.
+			set each_prefix (string sub -l $path_prefix_length -- $common_prefix)
+			set common_prefix_length (math $common_prefix_length - $path_prefix_length)
+			set common_prefix (string sub -s $new_start -- $common_prefix)
+		end
+
+		# Detect whether descriptions are present and the length of each completion.
+		set -l has_descriptions false
+		set -l longest_completion $_fzf_complete_min_description_offset
+		for i in (seq (count $completions))
+			if string match --quiet -- "*"\t"*" $completions[$i]
+				set has_descriptions true
+
+				# Here we additionally remember the longest completion to align the descriptions in fzf later
+				set longest_completion (math max $longest_completion,(string length -- $actual_completions[$i]))
+				set completions[$i] $actual_completions[$i]\t(set_color -i yellow)$descriptions[$i](set_color normal)
+			end
+		end
+
+		# FIXME Would technically work, but not sure it's worth the effort to match the descriptions
+		# after filtering.
+		# Remove tokens from the completion list that are already present on the current commandline.
+		#set -l all_tokens (commandline -o)
+		#set -l current_token_index (math 1 + (count (commandline -co)))
+		#set --erase all_tokens[$current_token_index]
+		#set -l remaining (comm --zero-terminated -23 (string join0 -- $actual_completions | psub) (string join0 -- $all_tokens | sort | psub) | string split0)
+
+		# TODO pressing / in a completion should add the completion and immediately start a new completion
+		test $has_descriptions = true; and set -l fzf_complete_description_opts \
+			--tabstop=(math 2 + $longest_completion)
+		set -l fzf_output (
+			string join0 -- $completions \
+			| _fzf_wrapper \
+				--read0 \
+				--print0 \
+				--ansi \
+				--multi \
+				--bind=tab:down,btab:up,change:top,ctrl-space:toggle \
+				--tiebreak=begin \
+				--query=$common_prefix \
+				--print-query \
+				$fzf_complete_description_opts \
+				$fzf_complete_opts \
+			| cut -f1 --zero-terminated \
+			| string split0)
+
+		switch $pipestatus[2]
+			case 0
+				# If something was selected, discard the current query
+				set results $fzf_output[2..-1]
+			case 1
+				# User accepted without selecting anything, thus we will
+				# use just the current query
+				set results $fzf_output[1]
+			case '*'
+				# Fzf failed, do nothing.
+				commandline -f repaint
+				return
+		end
+	end
+
+	set -l prefix ""
+	set -l suffix " "
+	# By default we want to append a space to completed options.
+	# While this is always true when competing multiple things, there
+	# are some cases in which we don't want to add a space:
+	if test (count $results) -eq 1
+		# When a completion ends in a / we usually want to keep adding to that completion.
+		# This may be because it is a directory (or link to a directory), or just
+		# an option that takes a category/name tuple like `emerge app-shells/fish`.
+		# Also, completions like ~something are directories.
+		#
+		# In a similar manner, completions ending in = are usually part of an argument
+		# that expects a parameter (like --color= or dd if=). The same logic applies here.
+		set -l first_char (string sub -l 1 -- $results[1])
+		set -l last_char (string sub -s -1 -- $results[1])
+		if test $last_char = =; or test $last_char = /
+			set suffix ""
+		else if string match --regex --quiet -- '^~[^/]*$' $results[1]
+			set suffix "/"
+		end
+	end
+
+	if not contains -- "--" (commandline -co)
+		# If a path is being completed and it starts with --, we add -- to terminate argument interpreting.
+		# Technically not always correct (the program may not accept --), but the alternative is worse
+		# and this will make the user notice it.
+		for r in $results
+			test - = (string sub -l 1 -- $r); and test -e $r; and set prefix "-- "; and break
+		end
+	end
+
+	# If each_prefix is set, we need to apply it to each result
+	# before replacing the token
+	if test -n $each_prefix
+		echo $each_prefix
+		set -l new_tokens
+		for r in $results
+			set -a new_tokens $each_prefix$r
+		end
+		commandline -t -- "$prefix$new_tokens$suffix"
+	else
+		commandline -t -- "$prefix$results$suffix"
+	end
+
+	commandline -f repaint
+end
--- a/functions/_fzf_search_completions.fish
+++ b/functions/_fzf_search_completions.fish
@@ -10,7 +10,7 @@ set --global _fzf_search_completions_min_description_offset 14
 function _fzf_search_completions --description "Shell completion using fzf"
 	# Produce a list of completions from which to choose
 	# and do nothing if there is nothing to select from
-	set -l cmd (commandline -co) (commandline -t)
+	set -l cmd (commandline --cut-at-cursor --tokenize) (commandline --current-token)
 	set -l completions (complete --escape --do-complete "$cmd")
 	test (count $completions) -eq 0; and return
 
--- a/functions/_fzf_search_completions.fish
+++ b/functions/_fzf_search_completions.fish
@@ -3,7 +3,10 @@
 #   (It thinks that it has to complete from / because `a_folder=/` looks like an argument to fish.
 # - fish cannot give reliable context information on completions. Knowing whether a completion
 #   is a file or argument can only be determined via a heuristic.
-set --global _fzf_complete_min_description_offset 14
+# - Completion is slow. Not because this script is slow but because fish's `complete` command
+#   can take several seconds even for just 100 entries. Just run `time complete --escape --do-complete l`
+#   to see for yourself.
+set --global _fzf_search_completions_min_description_offset 14
 function _fzf_search_completions --description "Shell completion using fzf"
 	# Produce a list of completions from which to choose
 	# and do nothing if there is nothing to select from
@@ -12,7 +15,7 @@ function _fzf_search_completions --description "Shell completion using fzf"
 	test (count $completions) -eq 0; and return
 
 	set -l results
-	set -l first_completion (printf "%s" $completions[1] | cut -f1 --zero-terminated | string split0)
+	set -l first_completion (string split --fields 1 --max 1 --right \t -- $completions[1])
 
 	# A prefix that will be prepended before each selected completion
 	set -l each_prefix ""
@@ -22,15 +25,20 @@ function _fzf_search_completions --description "Shell completion using fzf"
 	# want to review first (e.g. man ima<TAB> might match `feh`, an image viewer)
 	if test (count $completions) -eq 1; and test (string sub -l (string length -- (commandline -ct)) -- $first_completion) = (commandline -ct)
 		# If there is only one option then we don't need fzf
-		# Everytime we need to cut, we need zero termination to
-		# avoid newlines from being interpreted later.
 		set results $first_completion
 	else
-		# We preprocess the whole list to prevent spawning a slow cut process for each completion.
-		# From here on, make sure to use -- and zero termination everywhere to prevent introducing
+		# Preprocess the whole list and extract the actual completion any the corresponding description (if any)
+		# From here on, make sure to use -- (argument list end) and zero termination everywhere to prevent introducing
 		# new edge-cases where characters are interpreted wrongly
-		set -l actual_completions (string join0 -- $completions | cut -f1 --zero-terminated | string split0)
-		set -l descriptions (string join0 -- $completions | cut -f2- --zero-terminated | string split0)
+		set -l actual_completions
+		set -l descriptions
+		for i in (seq (count $completions))
+			# split on \t from the right, at most one time and use the first (left) field.
+			# This is the actual completion
+			set actual_completions[$i] (string split --fields 1 --max 1 --right \t -- $completions[$i])
+			# The other field is the description, if it exists (otherwise set empty).
+			set descriptions[$i] (string split --fields 2 --max 1 --right \t -- $completions[$i] ; or echo "")
+		end
 
 		# Find the common prefix of all completions. Yes this seems to be a hard
 		# thing to do in fish if it should be fast. No external process may be
@@ -77,7 +85,7 @@ function _fzf_search_completions --description "Shell completion using fzf"
 
 		# Detect whether descriptions are present and the length of each completion.
 		set -l has_descriptions false
-		set -l longest_completion $_fzf_complete_min_description_offset
+		set -l longest_completion $_fzf_search_completions_min_description_offset
 		for i in (seq (count $completions))
 			if string match --quiet -- "*"\t"*" $completions[$i]
 				set has_descriptions true
@@ -112,7 +120,6 @@ function _fzf_search_completions --description "Shell completion using fzf"
 				--print-query \
 				$fzf_complete_description_opts \
 				$fzf_complete_opts \
-			| cut -f1 --zero-terminated \
 			| string split0)
 
 		switch $pipestatus[2]
@@ -127,6 +134,11 @@ function _fzf_search_completions --description "Shell completion using fzf"
 				# Fzf failed, do nothing.
 				commandline -f repaint
 				return
+		end
+
+		# Strip anything after last \t (the descriptions)
+		for i in (seq (count $results))
+			set results[$i] (string split --fields 1 --max 1 --right \t -- $results[$i])
 		end
 	end
 
@@ -164,7 +176,6 @@ function _fzf_search_completions --description "Shell completion using fzf"
 	# If each_prefix is set, we need to apply it to each result
 	# before replacing the token
 	if test -n $each_prefix
-		echo $each_prefix
 		set -l new_tokens
 		for r in $results
 			set -a new_tokens $each_prefix$r
--- a/functions/_fzf_search_completions.fish
+++ b/functions/_fzf_search_completions.fish
@@ -4,7 +4,7 @@
 # - fish cannot give reliable context information on completions. Knowing whether a completion
 #   is a file or argument can only be determined via a heuristic.
 set --global _fzf_complete_min_description_offset 14
-function _fzf_complete --description "Shell completion using fzf"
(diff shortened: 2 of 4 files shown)


PatrickF1:
I'm currently swamped but didn't want you to leave you hanging for too long so I'll give you my preliminary thoughts while not having throughly understood the code.

First off, I'm incredibly humbled and grateful you'd submit such a big contribution to fzf.fish. You seem to have really done the research and know what I'm looking for--good documented code, robustness, reliability, even wrote out detailed explanations so I can understand it faster. Just because of how busy in life I've become (new parent, new job, bad health circumstances) and how I've already fulfilled the goals I started out with when I ran `git init` on fzf.fish, I'm leaning towards not merging this. HOWEVER, if you are willing to get on a call to teach me the code AND stay on as a maintainer to maintain this particular feature, then I would be open to it.

oddlama:
> Just because of how busy in life I've become (new parent, new job, bad health circumstances) and how I've already fulfilled the goals I started out with when I ran `git init` on fzf.fish, I'm leaning towards not merging this.
> HOWEVER, if you are willing to get on a call to teach me the code AND stay on as a maintainer to maintain this particular feature, then I would be open to it.

No worries, I can totally understand that. I just already had done the work and figured it cannot hurt to contribute it. At this point though I think I have to politely decline your kind offer, since I am not sure whether I will stick with fish over a longer timespan. While fish is awesome in many ways, zsh's completions system is just an order of magnitude more powerful - and (un)fortunately I really accommodated myself to those features.

While I don't think this feature would require much maintenance (as long as no new features/config options are added), I fear I wouldn't do it justice in case I left fish again. I'm happy regardless of whether this is merged or not, but I really think a lot of people would like this feature - so let me propose an alternative:

- You could leave this PR open for now and add a short notice in the README that the feature is theoretically fully implemented but in search of a maintainer. This way people would notice it and maybe someone is willing to add the required tests, add a section to the README and maybe add some requested configuration options later.
- I'll hereby license the code in this PR under the MIT License, so if anyone want's to create a new project from it they may do so with attribution. Additionally, I'll allow you to distribute it under your original LICENSE if anyone want's to maintain it here.
- If any problems occur in the future, I'd of course be happy to help. Just @ me in the issues.

Oh and regardless of the above, I'm of course open to giving you a walk-through of the code if you like :)
And best of luck with parenting and your job!

PatrickF1:
> I don't think this feature would require much maintenance (as long as no new features/config options are added

That's what previous contributors told me but bugs or weird edge cases that appear to be bugs w/o a deep understanding of the implementation always crop up. 

Okay I understand as well from your perspective. I like your recommendation. It just seems unlikely to work so I'll probably either merge this or close this after I have enough time to study the code more deeply. I really hope you can make this your own plugin (I'd install it!) but OSS is masochism so I don't recommend it to anyone in good conscience.

oddlama:
> That's what previous contributors told me but bugs or weird edge cases that appear to be bugs w/o a deep understanding of the implementation always crop up.

True.
 
> Okay I understand as well from your perspective. I like your recommendation. It just seems unlikely to work so I'll probably either merge this or close this after I have enough time to study the code more deeply. I really hope you can make this your own plugin (I'd install it!) but OSS is masochism so I don't recommend it to anyone in good conscience.

This month I'm a little busy with other stuff too but around March to April I think I'll can have another look at it. If it hasn't been picked up by anyone until then I'll see if I can quickly package it as a separate plugin. Thanks for the insightful conversation :)

functions/fzf_configure_bindings.fish
```diff
@@ -16,13 +16,14 @@ function fzf_configure_bindings --description "Installs the default key bindings
...
         # index 1 = directory, 2 = git_log, 3 = git_status, 4 = history, 5 = processes, 6 = variables
-        set key_sequences \e\cf \e\cl \e\cs \cr \e\cp \cv # \c = control, \e = escape
+        set key_sequences \e\cf \e\cl \e\cs \cr \e\cp \cv \t # \c = control, \e = escape
         set --query _flag_directory && set key_sequences[1] "$_flag_directory"
         set --query _flag_git_log && set key_sequences[2] "$_flag_git_log"
         set --query _flag_git_status && set key_sequences[3] "$_flag_git_status"
         set --query _flag_history && set key_sequences[4] "$_flag_history"
         set --query _flag_processes && set key_sequences[5] "$_flag_processes"
         set --query _flag_variables && set key_sequences[6] "$_flag_variables"
+        set --query _flag_complete && set key_sequences[7] "$_flag_complete"
```
PatrickF1, line 26: $_flag_complete doesn't get set unless you modify options_spec above to add something like 'complete=?'

functions/_fzf_complete.fish
```diff
@@ -0,0 +1,178 @@
+# External limitations:
```
PatrickF1, line 1: to keep with the naming convention of the other search functions, this should be named "_fzf_search_completions.fish"

PatrickF1:
Installed this PR but the completions don't trigger at all. When I run bind, \t _fzf_complete doesn't even show up. Can you look into it please?

functions/_fzf_complete.fish
oddlama: Good point, all occurrences have been renamed accordingly

functions/fzf_configure_bindings.fish
oddlama: Thanks, I've added a new option accordingly

PatrickF1:
Thanks for the fixes! Oh boy, I was afraid of this: I think my cut on my machine is different than the cut on your machine:
```
set: given 1 indexes but 0 values
math: Error: Too few arguments
'2 +'
   ^
cut: illegal option -- -
```
tons of those errors no matter which command's options I'm trying to complete. Are you on Linux? I'm on Mac.

oddlama:
> tons of those errors no matter which command's options I'm trying to complete. Are you on Linux? I'm on Mac.
> `cut: illegal option -- -`

Yep I'm on NixOS. `cut` is the only required external utility (apart from fzf of course), but it is really essential. I already tried to not use it, but there's nothing native in fish that is fast enough to replace it.

I was wondering what other fish fzf plugins did, but it seems like they all rely on `cut` in the same way. The difference seems to be the `--zero-terminated`, which is not supported on the decades version that ships with mac (still based on the old NetBSD cut I presume). It is a crucial component that makes all of this robust, but fortunately according to [this stackexchange post](https://apple.stackexchange.com/a/268353) the solution is to just install `coreutils` to get the proper coreutils and not the dumbed-down version. Can't check though, I don't own any macs.

PatrickF1:
Oh noooo that's really unfortunate. Not only do I not want to introduce a new dependency, the main problem I foresee is that if we add in Search Completions, either
- we turn it on automatically and users who don't have coreutils cut encounter their tab completions being broken for no explicable reason
- we keep it off and users who don't regularly check fzf.fish for updates have no idea that it's now available

Unfortunately, with fisher, I have very limited support for communicating with the users. The best I can do is insert something on the plugin update and how they see it, which historically has produced mixed results. Maybe this does need to be a new plugin...
That said, I got it working and it's quite lovely.

roland-5:
How can I use this? I downloaded changes from this PR to right places and it is for searching files (Ctrl+Alt+F)?

PatrickF1:
Down the changes, run fisher install, and it should be triggered by tab

roland-5:
Thank you. Damn, this looks nice. If problem is with Mac users only, so maybe make this as a plugin? I will use this very often from now on and I think there would be more people happy with this. It seems better option than not merging this.

PatrickF1:
If I had more time, I’d look into seeing if the cut logic can be emulated
using the fish string combination and fzf’s field expression functionality.
Can someone who is reading this please take a look? I and many others will
be eternally grateful!

oddlama:
> If I had more time, I’d look into seeing if the cut logic can be emulated
using the fish string combination and fzf’s field expression functionality.

I was wrong before, I just had an idea how to realize this. Will push an updated version soon.

And just a reminder if completions seem very slow (such as `l<TAB>`), it's not because the script is slow, but unfortunately because fish's completion system is just hellishly slow (just time the raw command: `time complete --escape --do-complete l`).

PatrickF1:
Nice @oddlama! So changing it from cut to `string` calls didn't slow it down significantly?

functions/_fzf_search_completions.fish
```diff
@@ -22,15 +25,20 @@ function _fzf_search_completions --description "Shell completion using fzf"
...
-		set -l descriptions (string join0 -- $completions | cut -f2- --zero-terminated | string split0)
+		set -l actual_completions
+		set -l descriptions
+		for i in (seq (count $completions))
+			# split on \t from the right, at most one time and use the first (left) field.
+			# This is the actual completion
+			set actual_completions[$i] (string split --fields 1 --max 1 --right \t -- $completions[$i])
+			# The other field is the description, if it exists (otherwise set empty).
+			set descriptions[$i] (string split --fields 2 --max 1 --right \t -- $completions[$i] ; or echo "")
+		end
```
PatrickF1, line 41: you can actually do something like
```fish
for completion in $completions
    set --append actual (string split $completion)
    set --append description (string split $completion)
```
to make this part easier to read :)

Susensio:
How does this implementation differs from https://github.com/gazorby/fifc ?

juliavdkris:
I just wanted to say this is the most positive and mutually understanding/respectful interaction in OSS I've seen in a while. Much love to both of you <3

functions/_fzf_search_completions.fish
oddlama: Done

oddlama:
> So changing it from cut to string calls didn't slow it down significantly?

No not noticeably. The main delay seems to come from fish itself, so nothing we can influence.

Oh and I just noticed that this will bump the minimum required fish version to 3.4.0 (released 12.03.2022) from 3.2.0, because of [complete --escape](https://github.com/fish-shell/fish-shell/blob/8c4bbe89e10217e1ebec782d654fc59008c6e0ee/CHANGELOG.rst?plain=1#L551). Just another thing to be aware of.

roland-5:
I didn't notice any difference in time respond if I do tab without any letter (where everything will show off) on my Orange Pi 3 with Debian Sid. It take little time, but the same amount (for OPi3) in previous and actual implementation.
On my main computer I didn't even see any lag with tab completion.

PatrickF1:
Hi again, I honestly don't have my life together now and it's going to be a while before I do. It doesn't help that the code is very complex so it's taking me a long time to digest. That said, I'm leaning towards merging this and would appreciate your patience.

Since I'm only digesting a bit of your code at a time, do you mind if I leave piece meal feedback on your PR? I know it's annoying that you don't get it all at once to address in one fell swoop but that's the best I can do.

oddlama:
Sure, take all the time you need to sort out your personal stuff first! I'll just address your comments as you go along.

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+# - fish cannot give reliable context information on completions. Knowing whether a completion
+#   is a file or argument can only be determined via a heuristic.
+# - Completion is slow. Not because this script is slow but because fish's `complete` command
+#   can take several seconds even for just 100 entries. Just run `time complete --escape --do-complete l`
+#   to see for yourself.
+set --global _fzf_search_completions_min_description_offset 14
+function _fzf_search_completions --description "Shell completion using fzf"
+	# Produce a list of completions from which to choose
+	# and do nothing if there is nothing to select from
+	set -l cmd (commandline -co) (commandline -t)
```
PatrickF1, line 13: could you use the long version of these flags please? `--cut-at-cursor --tokenize`, `--current-token`. Btw, how come you need the second `(commandline -t)`?

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+	set -l completions (complete --escape --do-complete "$cmd")
+	test (count $completions) -eq 0; and return
```
PatrickF1, line 15: this would read better (and be slightly more performant) as `test -z $completions; and return`. Might need `--`, not sure

PatrickF1:
You mention in the PR "with less bugs than fish's internal pager". What bugs does the fish internal completion pager suffer from that this doesn't?

functions/_fzf_search_completions.fish
oddlama: Good suggestion, will do. Regarding your question, consider `ls -l --color=<TAB>`
- `commandline --cut-at-cursor --tokenize` will yield `ls` and `-l` (only tokens before the cursor)
- `commandline --current-token` yields `--color=` (the current token until the cursor)

functions/_fzf_search_completions.fish
oddlama: I have to disagree here, since `test -z $completions` actually has different semantics and that would obscure the original intention (which actually is to test whether the array has any members).

You would also need to add quotes like `test -z "$completions"` to prevent passing the array entries as separate parameters `test -z entry1 entry2` is an invalid command, but then there is this edge case: `set arr ""; test -z "$arr"` is true even though the array is not actually empty.

oddlama:
> You mention in the PR "with less bugs than fish's internal pager". What bugs does the fish internal completion pager suffer from that this doesn't?

Now that you mention it, i think I was wrong. Apparently this is not true anymore since fish 3.4, before which the internal pager was not able to complete filenames with newlines (or other weird special characters) in them, which does work now. Since this PR raises the requirement for fish to 3.4 it would be very unfair to say it has less bugs, instead I just failed to make the connection that the introduction of `complete --escape` also fixed a lot of things for fish's internal pager. Sorry for that, should've checked more carefully.

(But don't worry, there still are a lot of bugs in fish's completion system that we inherit 😢 , e.g. try to do `touch aaa\nbbb` then `ls a<TAB>` works but `dd if=<TAB>` doesn't - I guess at least we also inherit the fixes automatically, when fish decides to fix that)

PatrickF1:
Ok so correctness wise, you think this PR is on par with fish's internal pager, and the improvement on top of it is purely having fzf as the interface?

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+# - fish cannot give reliable context information on completions. Knowing whether a completion
+#   is a file or argument can only be determined via a heuristic.
+# - Completion is slow. Not because this script is slow but because fish's `complete` command
+#   can take several seconds even for just 100 entries. Just run `time complete --escape --do-complete l`
+#   to see for yourself.
+set --global _fzf_search_completions_min_description_offset 14
+function _fzf_search_completions --description "Shell completion using fzf"
+	# Produce a list of completions from which to choose
+	# and do nothing if there is nothing to select from
+	set -l cmd (commandline --cut-at-cursor --tokenize) (commandline --current-token)
```
PatrickF1, line 13: could you explain why you do this thing of breaking up the command into two parts (before cursor and the current token)? is the reason because you want to remove everything after the current token?

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+	set -l completions (complete --escape --do-complete "$cmd")
+	test (count $completions) -eq 0; and return
+
+	set -l results
+	set -l first_completion (string split --fields 1 --max 1 --right \t -- $completions[1])
+
+	# A prefix that will be prepended before each selected completion
+	set -l each_prefix ""
+
+	# Only skip fzf if there is a single completion and it starts with the expected completion prefix.
```
PatrickF1, line 23: WDYM "starts with the expected completion prefix"? why would `complete --do-complete` ever return a completion the user wouldn't want?

functions/fzf_configure_bindings.fish
```diff
@@ -4,7 +4,7 @@ function fzf_configure_bindings --description "Installs the default key bindings
     # no need to install bindings if not in interactive mode or running tests
     status is-interactive || test "$CI" = true; or return
 
-    set options_spec h/help 'directory=?' 'git_log=?' 'git_status=?' 'history=?' 'processes=?' 'variables=?'
+    set options_spec h/help 'directory=?' 'git_log=?' 'git_status=?' 'history=?' 'processes=?' 'variables=?' 'completions=?'
```
PatrickF1, line 7: reminder to self (or I suppose you could do it too oddlama), need to update the configure bindings help function.

functions/_fzf_search_completions.fish
oddlama: Yes, for generating completions we only want to consider what comes before the cursor. The reason why keep it in an array using the --tokenize flag is that this still allows the code later to access the individual tokens, instead of coercing it to a single string. This is currently not needed technically, but the commented out part below might use it in the future to eliminate completions that are already present (eg. passing `--color` twice makes no sense).

functions/_fzf_search_completions.fish
oddlama: Sometimes the shell transforms or eliminates quotes, variables, ...
If that happens, I don't want this to silently change the commandline and instead prompt the user with fzf so any change is previewed and can be cancelled if needed.

Consider `ls foo"bar"` could be completed as `foobar`, but the user might have intended to place quotes there to e.g. later insert content with spaces more easily. I often do that and it's annoying if <TAB> deletes the quotes.

functions/_fzf_search_completions.fish
PatrickF1: I see thanks for explaining. 

> This is currently not needed technically, but the commented out part below might use it in the future to eliminate completions that are already present (eg. passing --color twice makes no sense).

For the sake of a lower-risk and faster MVP, can you update this to just use normal `complete --do-complete`? I also suspect we'll never build this out because even though for most commands the same option twice doesn't make sense, there will be a few in the wild that breaks that assumption.

functions/_fzf_search_completions.fish
PatrickF1: Ah okay this is consistent with how fish currently operates: if only one completion, it will auto-complete while preserving existing quotes and backslashes. Makes sense, good thinking!

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+	# Otherwise, the completion originates from a match in the description which the user might
+	# want to review first (e.g. man ima<TAB> might match `feh`, an image viewer)
```
PatrickF1, line 25: I don't think this is possible? `complete --do-complete "fzf_configure_bindings change"` and `complete --do-complete "fzf_configure_bindings -change"` both return nothing even though `change` matches the description of several `fzf_configure_bindings` options

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+		set -l descriptions
+		for i in (seq (count $completions))
+			# split on \t from the right, at most one time and use the first (left) field.
+			# This is the actual completion
+			set --append actual_completions (string split --fields 1 --max 1 --right \t -- $completions[$i])
+			# The other field is the description, if it exists (otherwise set empty).
+			set --append descriptions (string split --fields 2 --max 1 --right \t -- $completions[$i] ; or echo "")
+		end
+
+		# Find the common prefix of all completions. Yes this seems to be a hard
```
PatrickF1, line 43: Oh shoot, I just realized that by overriding the existing tab functionality, we actually need to fill in a lot more than just searching completions, including filling up to the common prefix. Hmm now I'm on the fence again about implementing this UNLESS we use a different key binding and shed this feature, so the user knows when they call the function, they are purely doing completions search and nothing else. I really do not want to be in the business of replicating what fish already does, only to supplement it.

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+			| string split0)
+
+		switch $pipestatus[2]
+			case 0
+				# If something was selected, discard the current query
+				set results $fzf_output[2..-1]
+			case 1
+				# User accepted without selecting anything, thus we will
+				# use just the current query
+				set results $fzf_output[1]
```
PatrickF1, line 132: This comment is better described as "Nothing matched the user's query so just use their query".

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+		# When a completion ends in a / we usually want to keep adding to that completion.
+		# This may be because it is a directory (or link to a directory), or just
+		# an option that takes a category/name tuple like `emerge app-shells/fish`.
+		# Also, completions like ~something are directories.
+		#
+		# In a similar manner, completions ending in = are usually part of an argument
+		# that expects a parameter (like --color= or dd if=). The same logic applies here.
+		set -l first_char (string sub -l 1 -- $results[1])
+		set -l last_char (string sub -s -1 -- $results[1])
+		if test $last_char = =; or test $last_char = /
```
PatrickF1, line 160: you can use `-o` instead of `; or test`

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+			set suffix "/"
+		end
+	end
+
+	if not contains -- "--" (commandline -co)
+		# If a path is being completed and it starts with --, we add -- to terminate argument interpreting.
+		# Technically not always correct (the program may not accept --), but the alternative is worse
+		# and this will make the user notice it.
+		for r in $results
+			test - = (string sub -l 1 -- $r); and test -e $r; and set prefix "-- "; and break
```
PatrickF1, line 172: can you make this into a normal if statement for readability, please?
also can use `-a` to merge the two tests

functions/_fzf_search_completions.fish
```diff
@@ -0,0 +1,189 @@
...
+	set -l prefix ""
+	set -l suffix " "
+	# By default we want to append a space to completed options.
+	# While this is always true when competing multiple things, there
+	# are some cases in which we don't want to add a space:
+	if test (count $results) -eq 1
+		# When a completion ends in a / we usually want to keep adding to that completion.
+		# This may be because it is a directory (or link to a directory), or just
+		# an option that takes a category/name tuple like `emerge app-shells/fish`.
+		# Also, completions like ~something are directories.
```
PatrickF1, line 154: wdym `~something`?

PatrickF1:
Hi @oddlama, finally finished my first read of the code (sorry I really had almost no free time until today) and I probably understand about 85% of it. Wow this is extremely thorough and it is truly a labor of love! Implementing completions and emulating current fish completion behavior is such a monumental undertaking. Kudos to your diligence and creativity making this work.

I still foresee a 2-3 passes before we beta-release it (don't add it to the default bindings but announce it). Let me know if you don't want to work on this anymore.

functions/_fzf_search_completions.fish
oddlama: The `feh` example was a real issue I encountered. Not sure if there are other fields used here, but this is definitely necessary to prevent matching non-real matches. Not sure how fish works here internally.

functions/_fzf_search_completions.fish
oddlama: Do with this PR what you want :)

functions/_fzf_search_completions.fish
oddlama: Home directories, such as ~patrick for example

oddlama:
> Wow this is extremely thorough and it is truly a labor of love! Implementing completions and emulating current fish completion behavior is such a monumental undertaking. Kudos to your diligence and creativity making this work.

Thanks 😄 

> I still foresee a 2-3 passes before we beta-release it (don't add it to the default bindings but announce it). Let me know if you don't want to work on this anymore.

I'm unfortunately also a little short on time right now. If you have any major concerns I'm of course happy to address them later. 
Other than that it would probably simpler and faster if you just force push any of your desired style changes directly, without going through me. Those won't affect the functionality either way and then you can just do what fits your style best.

functions/_fzf_search_completions.fish
PatrickF1: Oh you mean ~/something then?

functions/_fzf_search_completions.fish
PatrickF1: Odd! I'll look into this more.

functions/_fzf_search_completions.fish
oddlama: No, ~/something is the folder called `something` in the home directory of the currently logged-in user, while ~something is the home directory of the user named `something`

PatrickF1: Oh shoot I didn't know this syntax existed!

passionsfrucht:
Hi!

Foremost, let me please thank you for the inspiring and wholesome conversation. Truly inspiring!

Unfortunately, I share the same problem as you two do; not enough free time to be able to maintain the code.
And in general just enough knowledge about fish etc. to hurt myself.
Just today I stumbled over <https://github.com/PatrickF1/fzf.fish> and later this PR, both together do scratch my itches perfectly.
Finding out that the just learned `delta` integration for `git log` did not work in this fork, I just rebased the patches and trying out the results as stored in <https://github.com/passionsfrucht/fzf.fish>.

I just fixed the resulting conflicts in the most straightforward manner, and all of them made sense in my opinion. However, as I said, I am not fully aware of what I am doing in fish, so I may have introduced some bugs.

Additionally, I'm not sure how to handle this, so it fits into your workflows, thus I just leave it as a link here and let you know about it.
I plan to use this rebased version for now until I find something blocking and plan to let you know then.

PatrickF1:
Thanks Steffen! I think it's clear I don't have enough time for this so what I'll do is I'll beta release this by not adding a key binding by default, and whoever wants can bind it themselves and play around with it. Once I'm confident all the bugs have been caught, then I'll release it.

simonm:
Eagerly awaiting this. Is there anything I can do to help?

functions/_fzf_search_completions.fish
PatrickF1: Scratch my suggestion about doing the completion of the entire commandline vs considering what comes before the cursor. We definitely want to only complete what comes before the cursor rather than at the end of the command line.

PatrickF1:
Dang, I'm sorry everyone. Long story short I'm not going to merge this and never will include this feature. I finally had a free night and gave it a good go for a couple of hours. What made me realize this feature is not a good fit is because
- fish's built in Shift+tab is almost as good as using fzf: it filters and even searches based on the completion description. Actually, you can argue it's even better because it has special logic for autoselecting file completions and other stuff (read the comments on the PR)
- there are _tons_ of edge cases to consider. In my short two hours, I found 3 very difficult edge cases (@oddlama actually tried to handle them in the PR and he probably handled it 99%, not sure) and can only imagine if I actually worked on this for another 10 hours I'd find numerous more.

My conviction is that searching completions is better left to the fish-shell devs as an built-in. Sorry for getting everyone's hopes up!

PR #298, opened 2023-06-25
[Search Processes] use ps binary unaliased

tamirzb:
We should use the ps binary unalised because
- there are 1.7k lines of `alias ps=` on Github ([source](https://github.com/search?q=%22alias+ps%3D%22&type=code)), most of which will either cause the ps command to fail or to print redundant, hard-to-parse information.
- this is consistent with how fd is called unaliased from Search Directory


functions/_fzf_search_processes.fish
```diff
@@ -1,17 +1,19 @@
...
+        $ps_cmd -A -opid,command | \
         _fzf_wrapper --multi \
                     --prompt="Search Processes> " \
                     --query (commandline --current-token) \
                     --ansi \
                     # first line outputted by ps is a header, so we need to mark it as so
                     --header-lines=1 \
                     # ps uses exit code 1 if the process was not found, in which case show an message explaining so
-                    --preview="ps -o '$ps_preview_fmt' -p {1} || echo 'Cannot preview {1} because it exited.'" \
+                    --preview="$ps_cmd -ww -o '$ps_preview_fmt' -p {1} || echo 'Cannot preview {1} because it exited.'" \
```
PatrickF1, line 16: Is -w an option available on the default ps that ships with Linux, macOS, and whatever WIndows people use (sorry I'm rusty)?

tamirzb: I'm not that sure about Windows, do they have their own version of ps?

Anyway for Linux I can say that it works well on my machine (Arch Linux), and also the man page has this (https://man7.org/linux/man-pages/man1/ps.1.html).
For Mac I don't have one to try, but the man page seems to also have it (https://ss64.com/osx/ps.html).

PatrickF1:
For -ww, I don't get it. I checked out this PR and used it and don't see a difference. Also, in the docs, it says "When output is not to a terminal, an unlimited number of columns are always used." Right now, ps is not outputting to a tty, so I think -w might be irrelevant. Do you have screenshots of before and after adding -ww?

tamirzb:
Yeah I thought it was a bit odd that it doesn't work for me, especially since you did add `wrap` to fzf's `--preview-window` so I figured it should work for you, but for some reason it doesn't work for me.

Basically this is what I'm getting right now:
without_ww

And this is how it looks for me after my patch:
with_ww

Just a guess, but maybe it has to do with the fzf version? Like maybe at some point fzf was changed to make programs running under preview think they are outputting to a tty? My fzf version is 0.42.0.

Anyway to be honest it's not a big deal for me, so if you'd rather not merge this it's also ok.

PatrickF1:
> Just a guess, but maybe it has to do with the fzf version? Like maybe at some point fzf was changed to make programs running under preview think they are outputting to a tty? My fzf version is 0.42.0.

Hmm mine is also version 0.42.0. This is very odd. I surmise this is due to something specific to your setup so I'd rather not merge.

However, we can at least consider the part to force using the binary instead of the alias. Of course, there's pros and cos. Maybe some people want to be able to have the alias layer to control the binary. What's your use case for the alias and why you don't want the alias for fzf?

PatrickF1:
Bump on my question :)

tamirzb:
I mean there's not much to it. I have an alias for ps that is intended to prettify the output a bit but it ends up messing a bit with this code that parses ps output. I put it in a PR mainly because I saw you are doing the same with `fd` so thought you might want consistency.

Anyway, against it's really no big deal. If you don't want to merge it it's completely fine, I'll figure out some workaround for my config.

PatrickF1:
I'm probably going to merge this but right now I'm focused on building empathy with my users and learning about their use cases, how they use their tools in the wild. In light of that, do you mind sharing your ps alias?

PatrickF1 (approved):

PR #299, opened 2023-06-25
[Search history] Make date time format configurable

tamirzb:
Enable configuring the format of command execution times in Search History through the new `fzf_history_time_format` variable. This is highly valuable for non-US users who may be accustomed to reading dates as day-month (rather than month-day as the default is), or for people who want to search through previous years of command history.

### DIFF
--- a/functions/_fzf_search_history.fish
+++ b/functions/_fzf_search_history.fish
@@ -5,22 +5,26 @@ function _fzf_search_history --description "Search command history. Replace the 
         builtin history merge
     end
 
+    if not set --query fzf_history_time_format
+        # Reference https://devhints.io/strftime to understand strftime format symbols
+        set -f fzf_history_time_format "%m-%d %H:%M:%S"
+    end
+
     # Delinate commands throughout pipeline using null rather than newlines because commands can be multi-line
     set -f commands_selected (
-        # Reference https://devhints.io/strftime to understand strftime format symbols
-        builtin history --null --show-time="%m-%d %H:%M:%S │ " |
+        builtin history --null --show-time="$fzf_history_time_format │ " |
         _fzf_wrapper --read0 \
             --print0 \
             --multi \
             --tiebreak=index \
             --prompt="Search History> " \
             --query=(commandline) \
-            --preview="echo -- {4..} | fish_indent --ansi" \
+            --preview="echo -- {} | string replace --regex '^.*? │ ' '' | fish_indent --ansi" \
             --preview-window="bottom:3:wrap" \
             $fzf_history_opts |
         string split0 |
         # remove timestamps from commands selected
-        string replace --regex '^\d\d-\d\d \d\d:\d\d:\d\d │ ' ''
+        string replace --regex '^.*? │ ' ''
     )
 
     if test $status -eq 0
--- a/README.md
+++ b/README.md
@@ -179,6 +179,16 @@ set fzf_git_log_format "%H %s"
 
 The format must be one line per commit and the hash must be the first field, or else Search Git Log will fail to determine which commits you selected.
 
+### Change the time format used by Search History
+
+Search History prints the time each command was executed. To use your own [time format](https://devhints.io/strftime), set it in `fzf_history_time_format`. For example, this shows on what day of the week the command was executed:
+
+```fish
+set fzf_history_time_format "%A"
+```
+
+Note that the time format cannot include the unicode character `│` (not to be confused with the very similar `|`, the pipe symbol, which is allowed).
+
 ## Further reading
 
 Find answers to these questions and more in the [project Wiki](https://github.com/PatrickF1/fzf.fish/wiki):
--- a/README.md
+++ b/README.md
@@ -181,8 +181,8 @@ The format must be one line per commit and the hash must be the first field, or 
 
 ### Change the date time format used by Search History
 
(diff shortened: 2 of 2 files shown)


functions/_fzf_search_history.fish
```diff
@@ -20,7 +24,7 @@ function _fzf_search_history --description "Search command history. Replace the
             $fzf_history_opts |
         string split0 |
         # remove timestamps from commands selected
-        string replace --regex '^\d\d-\d\d \d\d:\d\d:\d\d │ ' ''
+        string replace --regex '^.* │ ' ''
```
PatrickF1, line 27: Is this greedy? Cause if so and the command itself contains a |, then it'll consume too much

tamirzb: No, I tested this with commands that had │and it worked correctly. Only way this can go wrong is if the user puts │ in their `fzf_history_time_format`, but I considered it a pretty far fetched edge case so didn't handle it.

If you prefer, I could add a check to make sure there is no │ in the user's `fzf_history_time_format`.

Alternatively, if you want to feel a bit safer about the regex, I can change the `*` to `[^│]`.

tamirzb: But I did just realize that I didn't handle it correctly for the preview window, so fixed that.

tamirzb:
Update: I realized that I didn't include the regex for the preview window, so fixed that.

functions/_fzf_search_history.fish
PatrickF1: > If you prefer, I could add a check to make sure there is no │ in the user's fzf_history_time_format.

Let's just put that in the docs. Can you take a stab at updating the readme? And I'll polish it for you.

> No, I tested this with commands that had │and it worked correctly.

Great, but now I'm confused why it works...?

tamirzb: Ok actually you are right, my bad, the regex is greedy. I will fix it. And yes I will also update the readme.

tamirzb: Ok, fixed the code and updated the README. Go ahead and polish it :)

functions/_fzf_search_history.fish
```diff
@@ -5,22 +5,26 @@ function _fzf_search_history --description "Search command history. Replace the
...
             --prompt="Search History> " \
             --query=(commandline) \
-            --preview="echo -- {4..} | fish_indent --ansi" \
+            --preview="echo -- {} | string replace --regex '^[^│]* │ ' '' | fish_indent --ansi" \
             --preview-window="bottom:3:wrap" \
             $fzf_history_opts |
         string split0 |
         # remove timestamps from commands selected
-        string replace --regex '^\d\d-\d\d \d\d:\d\d:\d\d │ ' ''
+        string replace --regex '^[^│]* │ ' ''
```
PatrickF1, line 27: Would you mind just doing the non-greedy version of what you had? I think it's `.*?`. I think it's more idiomatic and slightly easier to read

tamirzb: I realized I was wrong and it didn't work properly. When I had ` │ ` inside a command in my history it removed also the beginning of the command.

PatrickF1: Yeah but that's probably because you didn't make it non-greedy. That's what the ? does.

tamirzb: Ah ok sure, changed it.

PatrickF1:
Nice, looks good! I'll polish the readme tomorrow. Btw, what's your use case for configuring the date time format?

tamirzb:
I guess two things: First, my history spans over multiple years, so when I'm searching for things I can reach commands from different years, so having the year in there makes it much more insightful. Second, I am not American, so having the month before the day confuses me :sweat_smile:

PatrickF1:
> I guess two things: First, my history spans over multiple years, so when I'm searching for things I can reach commands from different years, so having the year in there makes it much more insightful.

Great use case!

> Second, I am not American, so having the month before the day confuses me 😅

Oh my gosh me and my American ignorance! Please excuse me. Could you explain what language you speak primarily and what dates in that language looks like?

tamirzb:
> Oh my gosh me and my American ignorance! Please excuse me. Could you explain what language you speak primarily and what dates in that language looks like?

Hahaha no worries. And to be honest I don't think it's a uniqueness of where I'm from, rather of the US. As you can see on [Wikipedia](https://en.wikipedia.org/wiki/Date_format_by_country), most of the world uses day-month-year as their primary date format, and month-day-year is pretty rare.

Anyway, as long as it's customizable it's really no big deal :)

roland-5:
You both sucks. The best format is year-month-day. :P I must say, it's nice feature. I don't have that much history, but there was few times when I struggle searched something similar within years.

PatrickF1:
> > Oh my gosh me and my American ignorance! Please excuse me. Could you explain what language you speak primarily and what dates in that language looks like?
> 
> Hahaha no worries. And to be honest I don't think it's a uniqueness of where I'm from, rather of the US. As you can see on [Wikipedia](https://en.wikipedia.org/wiki/Date_format_by_country), most of the world uses day-month-year as their primary date format, and month-day-year is pretty rare.
> 
> Anyway, as long as it's customizable it's really no big deal :)

Haha I totally forgot about that. What's worse is we're still not on the metric system 😭. Thanks for your input! I'll merge this soon.

PatrickF1:
Thank you for your contribution here. I consider it a valuable contribution for a large swath of fzf.fish users.

Btw, would you mind starring the repo now that you've contributed to it? Trying to get to 2k stars 😁

PR #302, opened 2023-07-31
[Search History][Search Git Log] use history scoring scheme

t-nil:
This improves ranking by using age not only as tiebreaker, but giving it more weight. I for example have some commands which I use almost daily but which always appear third or even lower in the list, due to some other commands (which I didn't use for months) matching *a little bit* better.

From the man page:

 ```
      --scheme=SCHEME
              Choose scoring scheme tailored for different types of input.

              default  Generic scoring scheme designed to work well with any type of input
              path     Scoring scheme for paths (additional bonus point only after path separator)
              history  Scoring scheme for command history (no additional bonus points).
                       Sets --tiebreak=index as well.
```

### DIFF
--- a/functions/_fzf_search_git_log.fish
+++ b/functions/_fzf_search_git_log.fish
@@ -10,7 +10,7 @@ function _fzf_search_git_log --description "Search the output of git log and pre
             git log --no-show-signature --color=always --format=format:$fzf_git_log_format --date=short | \
             _fzf_wrapper --ansi \
                 --multi \
-                --tiebreak=index \
+                --scheme=history \
                 --prompt="Search Git Log> " \
                 --preview='git show --color=always --stat --patch {1}' \
                 --query=(commandline --current-token) \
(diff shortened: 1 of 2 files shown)


xconverge:
@PatrickF1 this broke my system for some reason

I suspect the version of fzf in the repo for ubuntu is too old to support this

I get  ```unknown option: --scheme=history``` flashing really quickly when I do ctrl+r

Ubuntu 22.04 (LTS) is on fzf 0.29

I will try updating fzf manually I suppose

xconverge:
Yes building/installing fzf from source worked, but I do want you to be aware, it would have been nice to have this only apply in the wrapper if it was supported by the version

PatrickF1:
Oops my bad, yeah I forgot to bump the minimum fzf version and then I should have made this a major release. Thanks for the FYI

t-nil:
> Hi Florian, YESSS I was looking into this when the history scheme came out but I never got around to this. This is a change I want to make. Two things
> 
> 1. can you explain how this "Scoring scheme for command history (no additional bonus points)." means that more recent entries will have more weight?
> 2. can you make the change for git log too, please?

Hey Patrick,
sorry, I am only erratically active on github. At that time I didn't find out anything else except that the scheme seemed to work for my history search.
Thanks alot for taking the time, specifically fixing this and investigating on fzf Issues, and in general for maintaining this very useful set of scripts :)

PR #309, opened 2023-09-09
replace exa with eza in example now that exa is unmaintained

nicoandmee:
[exa](https://github.com/ogham/exa/issues/1243) is now unmaintained. `eza` is a fork that is essentially a drop in replacement for `exa`. This PR encourages its use over the former. 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Find more ideas and implementation tips in the [Cookbook](https://github.com/Pat
 To use your own directory preview command, set it in `fzf_preview_dir_cmd`:
 
 ```fish
-set fzf_preview_dir_cmd exa --all --color=always
+set fzf_preview_dir_cmd eza --all --color=always
 ```
 
 And to use your own file preview command, set it in `fzf_preview_file_cmd`:


roland-5:
Arch Linux migrated to `eza`, but point release distributions like Ubuntu, Debian etc. will not so fast. It would be better to add info about `eza`, without deleting `exa`, because of it.

PatrickF1:
Wow didn't know exa was unmaintained--been using it every day. Thanks for the suggestion.
@roland-5 are you saying that Ubuntu and Debian users would have trouble installing eza using the built in package manager?

roland-5:
If user of point release distribution don't use cargo/cargo-bin and only their distribution package manager, then for actual stable releases there isn't anything like `eza`, only still exists `exa`.
https://packages.debian.org/bookworm/exa Nor trixie (testing) or even sid (unstable) have eza package.
These distributions doesn't add new packages for stable releases so only security patches or rebuilds if necessary will be released for `exa` (if so). When Debian will ship eza (if someone will build it for sid and transfer it to trixie), then Ubuntu will suck it to their repositories. Probably Fedora, Open Suse and other point released distro will do the same.

PatrickF1:
Thanks for explaining the Linux world! Any guess on when Debian will pick up eza?

roland-5:
No idea. Sometimes they update or add some packages even before Arch Linux or Fedora Silverblue, but if there is no willing maintainer for specific application, it could be months or years.
If someone makes a package with Eza, in the case of Ubuntu, it would appear in the 24/04 release at the earliest. The easiest way would be to propose to install and use Eza as a developed application, and for those who are unable to do so, Exa.

roland-5:
Damn, there is no `starship` still in Debian/Ubuntu repos. :D

nicoandmee:
> Arch Linux migrated to `eza`, but point release distributions like Ubuntu, Debian etc. will not so fast. It would be better to add info about `eza`, without deleting `exa`, because of it.

I would agree, if it were going to affect actual functionality - this is more of a change in recommendation, orthogonal to the functionality of the plugin itself 😄

PatrickF1:
Yeah this has no actual effects on the plugin and it might be a long wait before Debian adopts eza. I'll give the deprecation of exa another week to marinate before I merge this.

PR #310, opened 2023-09-13
feat: search git-stash follows git-log

NextAlone:


PatrickF1:
Sorry I'm not open to this feature. Might change my mind later, but unlikely. I'm trying to keep down the feature set so there's less surface area to be maintained (and get PRs for).

NextAlone:
That's okay

